Issue #22, opened 2025-10-30
[repo agent] check dependencies in makefile


barney-s:
Currently in repo-agent/Makefile we do not check if the required tools are installed when running make.
For example kind, helm.

We need to fix it by adding check-prereqs or some similar target that checks for prerequisites as called out in repo-agent/Readme.md before running any other target in make.

barney-s:
fixed

PR #24, opened 2025-10-30
feat(repo-agent): add prerequisite checks to Makefile

barney-s:
Adds a 'check-prereqs' target to the repo-agent/Makefile to verify that 'kind', 'kubectl', and 'helm' are installed before running any other targets. This addresses issue #22.


aaron-prindle:
Nice! I hit a similar issue in not having `helm` in my first attempt to install.  

LGTM

PR #40, opened 2025-10-31
feat(repo-agent): record review feedback changes

### LINKED ISSUE #17: [repo agent] Add the ability to record review feedback changes made by user
When the user modifies the review output and uses it use that as a signal to finetune review prompts. 

One way to do it would be changing these parts:

1. When creating a Review entry in the review-api/main.go , add a new field in the redist HSET called agentDraft and set it to the same value as draft. Same for PR.
2. When processing  submitReview or submitIssueComment and the draft != agentDraft , we should create a new record entry in redis.
3. The record entry should be a HSET in redis under the key "hf:review:githubuser:$login:repo:$repo:pr:$pr" with fields "draft", "agentDraft", "prompt", "configdir" etc.
4. For issues something similar should be done under the key "hf:issue:githubuser:$login:repo:$repo:handler:$handler:pr:$pr" with fields "draft", "agentDraft", "prompt", "configdirname" etc.

barney-s:
This commit adds the ability to record review feedback changes made by the user. When a user modifies a review or an issue comment, the original agent-generated draft and the user's modified draft are stored in Redis. This data can be used for future analysis and fine-tuning of the review prompts.

Fixes #17

### DIFF
--- a/repo-agent/review-ui/review-api/main.go
+++ b/repo-agent/review-ui/review-api/main.go
@@ -423,6 +423,7 @@ func fetchAndPopulatePRs(ctx context.Context, repo string) {
 			"htmlurl", pr.HTMLURL,
 			"sandboxReplica", pr.SandboxReplica,
 			"draft", draft,
+			"agentDraft", draft,
 		).Err(); err != nil {
 			log.Printf("Failed to cache PR %s for repo %s: %v", pr.ID, repo, err)
 		}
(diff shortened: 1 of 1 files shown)


repo-agent/review-ui/review-api/main.go
```diff
@@ -472,6 +485,25 @@ func submitReview(c *gin.Context) {
...
 
+	if draft != agentDraft {
+		// Store feedback for fine-tuning
+		prompt, _, _ := unstructured.NestedString(repoWatch.Object, "spec", "review", "gemini", "prompt")
+		configdir, _, _ := unstructured.NestedString(repoWatch.Object, "spec", "review", "gemini", "configdirRef")
+		repoURL, _, _ := unstructured.NestedString(repoWatch.Object, "spec", "repoURL")
+		owner, _, _ := parseRepoURL(repoURL)
+
+		hfKey := fmt.Sprintf("hf:review:githubuser:%s:repo:%s:pr:%s", owner, repo, prID)
+		if err := rdb.HSet(ctx, hfKey,
```
aaron-prindle: Do we want to have the backend be configurable or have a split backend for redis/memory-items vs postgresql/storage-items?  I am not sure the final architecture but currently the delta feedback (important telemetry) would be wiped on reboots etc.

Is the assumption the redis instance will be long lived and always available?  My only concern would be that effectively we would to export this telemetry and hopefully across N teams and not sure if folks would understand that restarting their instance would wipe all telemetry, etc.

aaron-prindle:
LGTM

repo-agent/review-ui/review-api/main.go
barney-s: this is a great Q. I think this needs to be persistent across installs. Across cluster lifetimes.

PR #46, opened 2025-11-01
Changes to Install in GKE cluster

### LINKED ISSUE #44: [repo-agent] Install in GKE cluster
We need the make installer to be able to install in GKE cluster.
To that end we may have to do a bunch of things:

- [ ] dev/cloud scripts to setup GCP project, GKE cluster etc.
- [x] Build docker images and host them somewhere say GHCR. Done #15 
- [ ] make variable to Use GHCR images and deploy to cluster

may be more items.

barney-s:
* Update README with instructions
* Update dev scripts to support IMAGE_TAG
* Refactor make targets to support build + install in kind, vs install in any k8s cluster

Fixes #44

### DIFF
--- a/dev/tools/deploy-to-kube
+++ b/dev/tools/deploy-to-kube
@@ -91,7 +91,10 @@ def main(args):
 
             service_name = os.path.basename(root)
             image_prefix = utils.get_image_prefix(args)
-            image_tag = utils.get_image_tag()
+            if args.image_tag:
+                image_tag = args.image_tag
+            else:
+                image_tag = utils.get_image_tag()
 
             # Process each document in the file
             for doc in docs:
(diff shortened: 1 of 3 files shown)


dev/tools/deploy-to-kube
```diff
@@ -91,7 +91,10 @@ def main(args):
 
             service_name = os.path.basename(root)
             image_prefix = utils.get_image_prefix(args)
-            image_tag = utils.get_image_tag()
+            if args.image_tag:
```
justinsb: Not a blocker, but I'm wondering if we should have e.g. `args.image_tag()`, and if you specify `image_tag` then `image_tag()` returns that value.

`args` becomes more of a configuration object.

justinsb (approved):

PR #60, opened 2025-11-05
feat: github logic and namespace support

aaron-prindle:
This PR introduces full multi-tenancy support to the Repo Agent. It enables multiple users to log in via GitHub, with each user receiving an isolated workspace (Kubernetes namespace) for their review and issue sandboxes. It also includes a fallback "Guest" mode for unauthenticated access.

  * `gorilla` sessions
    *  We utilize gorilla/sessions with a cookie-based store to persist user authentication state securely across API requests. Upon successful GitHub OAuth login, the user's username is serialized into an encrypted session cookie. A dedicated middleware intercepts incoming requests, retrieving this session to identify the user and context-switch to their specific Kubernetes namespace. If the session is missing or invalid, the system uses the "guest" mode, defaulting operations to the default namespace. 

   * Backend (`review-ui/review-api`):
       * Implemented GitHub OAuth flow (/api/auth/* endpoints).
       * Added bootstrapNamespace logic to provision user namespaces, copy secrets/configmaps, and setup RBAC on-the-fly.
       * Updated all resource handlers (createRepoWatch, getPRs, etc.) to operate within the context of the authenticated user's namespace.
       * Added API endpoints to dynamically configure OAuth credentials from the UI.

   * Controller (`repowatch`):
       * Scoped ReviewSandbox and IssueSandbox listing to the RepoWatch namespace. This fixes a critical bug where sandboxes in one namespace blocked creation in others.

   * Frontend (`review-ui/ui`):
       * Added Login/Logout flow and user session state.
       * Added a configuration form to set GitHub Client ID/Secret if missing.
       * Restored the ability to provide a custom "Name" when adding a repository.

   * Infrastructure (`k8s/`, `Makefile`):
       * Updated api-rbac.yaml to grant the API permissions to manage Namespaces, ServiceAccounts, and RoleBindings.
       * Updated Makefile to support GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET environment variables for automated
         secret creation.
       * Added k8s/go-configmap-devcontainer.yaml to ensure the Go devcontainer config is available for bootstrapping.

NOTE: in a follow-up I plan on adding per namespace/user GITHUB_PAT support as well as per-repo GITHUB_PAT support. Currently all users use the same GITHUB_PAT.

updated login page:
image

(guest login, note no username)

image

(simultaneously running aaron-prindle Github login, not the username)
image

### DIFF
--- a/repo-agent/review-ui/review-api/main.go
+++ b/repo-agent/review-ui/review-api/main.go
@@ -231,6 +231,7 @@ func main() {
 	router.Use(RequestLoggerMiddleware())
 	router.Use(ResponseLoggerMiddleware())
 
+	// Public routes: Authentication flows and initial configuration do not require a session.
 	router.GET("/api/auth/login", authLogin)
 	router.GET("/api/auth/callback", authCallback)
 	router.GET("/api/auth/status", authStatus)
@@ -238,7 +239,7 @@ func main() {
 	router.GET("/api/auth/providers", getAuthProviders)
 	router.POST("/api/auth/github-config", updateGithubConfig)
 
-	// API routes
+	// Protected routes: The "api" group applies authMiddleware to enforce valid sessions for all downstream endpoints.
 	api := router.Group("/api")
 	api.Use(authMiddleware())
 	{


aaron-prindle:
Initial automated review

### Areas of Concern (Hacky or Problematic)

1.  **API Permissions (RBAC) Sprawl:** The API pod in `repo-agent-system` is now effectively a cluster admin. If this pod is compromised, the attacker controls the entire cluster.

    -   *Better approach:* Create a dedicated `ClusterRole` with only the narrow permissions needed (create NS, read specific secret in system namespace, create secrets in specific namespaces), rather than broad access to all resources.

1.  **Hardcoded Session Key:** In `main.go`:

```go
store := cookie.NewStore([]byte("secret"))
```

> This uses a hardcoded, weak secret for signing session cookies. In production, this would allow anyone to forge a session. It should be loaded from an environment variable (which gets loaded from a K8S Secret).

1.  **Concurrency & Latency in Bootstrapping:** The `bootstrapNamespace` function runs in the middleware on *every* request.

    -   *Performance concern:* While `k8sClientset....Get` is fast, hitting it on every single API call adds latency.
    -   *Race condition:* If multiple requests come in simultaneously for a new user, `bootstrapNamespace` might be executed multiple times in parallel, leading to race conditions in creating resources.

1.  **Secret Copying Pattern:** Copying secrets manually in Go code (`copySecret` function) is a bit fragile. If the original secret in `repo-agent-system` changes, the user secrets won't automatically update.

1.  **Makefile brittleness:** In `Makefile`:

```
kubectl get configmap go-devcontainer-json ... | sed 's/name: go-devcontainer-json/name: devcontainer-json/' | grep -v ...
```

> This looks brittle. It's better to have distinct YAML files in the `k8s/` or `examples/` folder rather than relying on `sed` to manipulate object names on the fly.

1.  **Error Handling in Bootstrap:** The `bootstrapNamespace` attempts to copy secrets, but if it fails (e.g., `copyConfigMap`), it just logs a debug/warning message and continues. This might result in a broken user namespace where the agent won't work, but the UI reports success.

1.  **Redis Keys Breaking Change:** Changing Redis keys from `pr:repo:...` to `pr:ns:%s:repo:%s...` will orphan all existing Redis data upon deployment. You might need a migration strategy if preserving existing data matters.

repo-agent/review-ui/review-api/main.go
```diff
@@ -87,356 +107,607 @@ func (w bodyLogWriter) Write(b []byte) (int, error) {
...
+	router.GET("/api/auth/callback", authCallback)
+	router.GET("/api/auth/status", authStatus)
+	router.POST("/api/auth/logout", authLogout)
+
 	api := router.Group("/api")
+	api.Use(authMiddleware())
 	{
 		api.GET("/repos", getRepos)
-		api.GET("/repo/:namespace/:repo/prs", getPRs)
-		api.POST("/repo/:namespace/:repo/prs/:id/draft", saveDraft)
```
barney-s: curious. If we remove namespace from the URL does this make the api-server single tenant ?

repo-agent/k8s/api-deployment.yaml
```diff
@@ -29,3 +29,15 @@ spec:
               name: github-token
               key: token
               optional: true # For local development
+        - name: GITHUB_CLIENT_ID
```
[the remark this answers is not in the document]
aaron-prindle: To use Github Oauth login you need to specify your GITHUB_CLIENT_ID and you GITHUB_CLIENT_SECRET for your Oauth app.  Related docs here:

https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authenticating-to-the-rest-api-with-an-oauth-app

barney-s:
This is a great step forward. Makes us think harder about MT.
Like we discussed offline, we need to support 2 installation paths. One for MT and one for ST.

MT installations would be like a managed service since we need to deal with service level oauth tokens vs user level PATs. 
ST would only use user PATs.

Also if possible would you be able to send the ST UI changes which takes a PAT and Gemini key and onboards the user first as a separate PR ? 

The second PR could be MT enabled with github/oauth signup.

aaron-prindle (commented):
bot comment - test

repo-agent/repowatch/controllers/review_system_prompt.go
```diff
@@ -73,14 +73,17 @@ additional review instructions:
 
 
 Output:
-- Output should be a valid YAML, and nothing else.
-- Each YAML output MUST be after a newline, with proper indent, and block scalar indicator ('|')
-- The output must be a YAML object of the type $PullRequestReviewRequest, according to the following Pydantic definitions:
+- The output MUST be valid YAML.
```
aaron-prindle: NOTE: I hit an issue where periodically gemini would not produce proper the proper formatted output so I updated the prompt to attempt to resolve this.  Lmk and I can split this out if preferred.

repo-agent/k8s/go-configmap-devcontainer.yaml
```diff
@@ -0,0 +1,22 @@
+# Configmap embedding devcontainer.json for VSCode Remote - Containers support
+apiVersion: v1
```
barney-s: use https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/repo-agent/k8s/configmap-devcontainer.yaml instead. 
rename that file and remove namespace maybe?

repo-agent/Makefile
```diff
@@ -19,6 +19,18 @@ ifndef GITHUB_PAT
 $(error GITHUB_PAT is not set. Please set it before running make.)
 endif
 
+ifndef ANTHROPIC_API_KEY
```
barney-s: We need to update the installer.sh to take in these envvars. 
And update the readme to show MT vs ST installation.,

repo-agent/review-ui/review-api/main.go
```diff
@@ -97,244 +116,605 @@ func (w bodyLogWriter) Write(b []byte) (int, error) {
...
 
-	// API routes
+	router.GET("/api/auth/login", authLogin)
+	router.GET("/api/auth/callback", authCallback)
+	router.GET("/api/auth/status", authStatus)
+	router.POST("/api/auth/logout", authLogout)
+	router.GET("/api/auth/providers", getAuthProviders)
+	router.POST("/api/auth/github-config", updateGithubConfig)
+
 	api := router.Group("/api")
```
barney-s: a comment on middleware usage and grouping of apis/

repo-agent/review-ui/review-api/main.go
```diff
@@ -376,240 +756,112 @@ func populateMockData() {
...
-			pr.SandboxReplica = prData["sandboxReplica"]
-		}
-		if _, ok := prData["review"]; ok {
-			pr.Review = prData["review"]
-		}
-		prs = append(prs, pr)
-	}
-	if err := iter.Err(); err != nil {
-		log.Printf("Error during Redis SCAN: %v", err)
+		prs = append(prs, PR{
```
barney-s: add a TODO to check for map keys before access

barney-s:
is this ready for merge @aaron-prindle  ?

repo-agent/k8s/go-configmap-devcontainer.yaml
aaron-prindle: I have changed things to now use configmap-devcontainer.yaml and deleted the additional go-configmap-devcontainer.yaml.

repo-agent/review-ui/review-api/main.go
aaron-prindle: Done, added two comments related to the different sections:
```go
	// Public routes: Authentication flows and initial configuration do not require a session.
...
	// Protected routes: The "api" group applies authMiddleware to enforce valid sessions for all downstream endpoints.
```

PR #63, opened 2025-11-06
Fix example to include llm info review section. 

barney-s:
Master has a bug. this fixes it. 

we need better defaulting. this is more of a quick fix for examples


barney-s:
@seans3 - FYI
@cheftako

seans3 (approved):
/lgtm

Looks like I missed two places where "llm" section should have been inserted. Sorry.

PR #66, opened 2025-11-06
Add docs on configdir-cli usage and support for including folder in path

barney-s:


repo-agent/README.md
```diff
@@ -156,6 +156,48 @@ The examples also demonstrate how to use `ConfigDir` and `devcontainer` to creat
...
 By customizing the `RepoWatch` resource and the `devcontainer` configuration, you can create a powerful and flexible Gemini agent that is tailored to your specific needs.
 
+#### Creating a configdir from your .gemini folder
+
+Build the `configdir-cli` binary
+```bash
+make build # build all the binaries
+```
+
+Create a `configdir` entry from your `.gemini` folder:
```
aaron-prindle: nit: currently this is framed higher level of usage (k8s admin) than I believe we want users to interact with.  Copying/mounting local files (.gemini, etc.) into the devcontainers should likely be something we support in the UI directly.  This is also likely a feature where each user want their own Gemini commands potentially or customization which would lend itself to a per-user (UI) feature.  Perhaps we can make a TODO to make this a first class feature (I don't know if in practice users will understand they can do this from the documentation here)

aaron-prindle:
/lgtm
/approve
left one nit related to created an issue around allowing UI support for mounting/copying local .gemini, etc. files

PR #96, opened 2025-11-11
Generalized LLM provider for review-sandbox

seans3:
* Moved `gemini-cli` LLM provider code in `review-sandbox` into a common directory (`pkg/llm`).
* Added `Provider` interface to generalize the backend LLM.
* Adds generalized `PostProcessor` functions to the LLM Provider interface
* `stripYAML` is now a post-processor function (added to provider in `review-sandbox/main.go`).
* Added unit tests to the LLM provider (coverage: 83.0%)
* Update go.mod/go.sum with `go mod tidy`.
```
$ go test -cover ./llm/...
ok  	github.com/gke-labs/gemini-for-kubernetes-development/repo-agent/pkg/llm	0.019s	coverage: 83.0% of statements
```

### DIFF
--- a/repo-agent/pkg/llm/provider.go
+++ b/repo-agent/pkg/llm/provider.go
@@ -14,7 +14,10 @@
 
 package llm
 
-import "fmt"
+import (
+	"bytes"
+	"fmt"
+)
 
 // PostProcessor defines the signature for functions that can post-process the LLM's raw output.
 type PostProcessor func([]byte) ([]byte, error)
@@ -31,10 +34,35 @@ type Provider interface {
 func NewLLMProvider(name string) (Provider, error) {
 	switch name {
 	case "gemini-cli":
-		g := &Gemini{Executor: &RealCommandExecutor{}}
-		g.AddPostProcessor(stripYAMLMarkers)
-		return g, nil
+		return &Gemini{Executor: &RealCommandExecutor{}}, nil
 	default:
 		return nil, fmt.Errorf("unknown provider: %s", name)
 	}
+}
+
+// StripYAMLMarkers looks for ```yaml and ``` markers in the input byte slice.
+// If found, it strips these markers and returns the content between them.
+// If markers are not found, the original byte slice is returned.
+func StripYAMLMarkers(input []byte) ([]byte, error) {
+	startMarker := []byte("```yaml")
+	endMarker := []byte("```")
+
+	startIndex := bytes.Index(input, startMarker)
+	if startIndex == -1 {
+		return input, nil // Start marker not found
+	}
+
+	// Adjust startIndex to point after the start marker
+	startIndex += len(startMarker)
+
+	endIndex := bytes.Index(input[startIndex:], endMarker)
+	if endIndex == -1 {
+		return input, nil // End marker not found after start marker
+	}
+
+	// Adjust endIndex to be relative to the original input slice
+	endIndex += startIndex
+
+	// Extract the content between the markers, trimming any leading/trailing whitespace
+	return bytes.TrimSpace(input[startIndex:endIndex]), nil
 }
--- a/repo-agent/review-sandbox/main.go
+++ b/repo-agent/review-sandbox/main.go
@@ -82,6 +82,7 @@ func runReview() error {
 	if err != nil {
 		return err
 	}
+	provider.AddPostProcessor(llm.StripYAMLMarkers)
 
 	if err := provider.Setup("/workspaces", "/tokens"); err != nil {
 		return err
--- a/repo-agent/pkg/llm/gemini.go
+++ b/repo-agent/pkg/llm/gemini.go
@@ -15,7 +15,6 @@
 package llm
 
 import (
-	"bytes"
 	"fmt"
 	"log"
 	"os"
@@ -86,31 +85,4 @@ func (g *Gemini) Run(agentPrompt string) ([]byte, error) {
 	}
 
 	return output, nil
-}
-
-// stripYAMLMarkers looks for ```yaml and ``` markers in the input byte slice.
-// If found, it strips these markers and returns the content between them.
-// If markers are not found, the original byte slice is returned.
-func stripYAMLMarkers(input []byte) ([]byte, error) {
-	startMarker := []byte("```yaml")
-	endMarker := []byte("```")
-
-	startIndex := bytes.Index(input, startMarker)
-	if startIndex == -1 {
-		return input, nil // Start marker not found
-	}
-
-	// Adjust startIndex to point after the start marker
-	startIndex += len(startMarker)
-
-	endIndex := bytes.Index(input[startIndex:], endMarker)
-	if endIndex == -1 {
-		return input, nil // End marker not found after start marker
-	}
-
-	// Adjust endIndex to be relative to the original input slice
-	endIndex += startIndex
-
-	// Extract the content between the markers, trimming any leading/trailing whitespace
-	return bytes.TrimSpace(input[startIndex:endIndex]), nil
 }
--- a/repo-agent/pkg/llm/gemini_test.go
+++ b/repo-agent/pkg/llm/gemini_test.go
@@ -23,53 +23,6 @@ import (
 	"testing"
 )
 
-func TestStripYAMLMarkers(t *testing.T) {
-	tests := []struct {
-		name    string
-		input   []byte
-		want    []byte
-		wantErr bool
-	}{
-		{
-			name:    "with markers",
-			input:   []byte("```yaml\nfoo: bar\n```"),
-			want:    []byte("foo: bar"),
-			wantErr: false,
-		},
-		{
-			name:    "without markers",
-			input:   []byte("foo: bar"),
-			want:    []byte("foo: bar"),
-			wantErr: false,
-		},
-		{
-			name:    "empty input",
-			input:   []byte(""),
-			want:    []byte(""),
-			wantErr: false,
-		},
-		{
-			name:    "only markers",
-			input:   []byte("```yaml\n```"),
-			want:    []byte(""),
-			wantErr: false,
-		},
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			got, err := stripYAMLMarkers(tt.input)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("stripYAMLMarkers() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
-			if !bytes.Equal(got, tt.want) {
-				t.Errorf("stripYAMLMarkers() = %q, want %q", got, tt.want)
-			}
-		})
-	}
-}
-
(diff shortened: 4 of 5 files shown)


repo-agent/review-sandbox/main.go
```diff
@@ -115,7 +114,7 @@ func runReview() error {
 			}
 		}
 
-		output, err := agentFn(currentPrompt)
+		output, err := provider.Run(currentPrompt)
```
barney-s: may be treat strip yaml as a processor to Run(). 
Reason being for some usecases (IssueHandlers) we expect text and no yaml.

repo-agent/llm/provider.go
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package llm
+
+import "fmt"
+
+type Provider interface {
+	Setup(workspacesDir, tokensDir string) error
+	Run(prompt string) ([]byte, error)
```
barney-s: Run can take processors that can do simple post processing. 
Processors can be processor([]byte)[]byte , error ?

repo-agent/llm/exec.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2025 Google LLC
+//
```
barney-s: not sure .. wondering if pkg/llm makes more sense ?

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+// Gemini is an Provider that uses the gemini-cli.
+//
+// Make sure that the Gemini struct implements the Provider interface.
+var _ Provider = &Gemini{}
+
```
barney-s: The `top_p` parameter is hardcoded to `0.95`. This is a key generation parameter that should be configurable to allow for tuning the model's output.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+type Gemini struct {
```
barney-s: The `top_k` parameter is hardcoded to `20`. This should be exposed as a configurable parameter.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+	Executor CommandExecutor
```
barney-s: The `temperature` is hardcoded to `1.0`. This is another critical parameter for controlling model creativity that should be configurable.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+}
```
barney-s: The `maxOutputTokens` is hardcoded. This should be configurable to allow users to adjust the response length based on their needs and the model's limits.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (g *Gemini) Setup(workspacesDir, tokensDir string) error {
+	// if .gemini directory exists in /workspaces copy it to home directory
+	geminiConfigDir := filepath.Join(workspacesDir, ".gemini")
+	if _, err := os.Stat(geminiConfigDir); err == nil {
+		log.Println(".gemini directory exists in /workspaces, copying to repo directory")
+		// if desitation .gemini directory exists move it to .gemini.bak
+		if _, err := os.Stat(".gemini"); err == nil {
+			log.Println(".gemini directory exists in repo directory, moving to .gemini.bak")
```
barney-s: The `gcloud` command path is hardcoded. This assumes `gcloud` is always in the system's `PATH`. To make this more robust, consider allowing the path to be configurable via an environment variable or a parameter in the `New` function.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+			err := os.Rename(".gemini", ".gemini.bak")
+			if err != nil {
+				return fmt.Errorf("failed to move .gemini to .gemini.bak: %v", err)
+			}
+		}
+		cmd := exec.Command("cp", "-R", geminiConfigDir, ".gemini")
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
```
barney-s: The error message "failed to get gcloud project" is a bit generic. It would be more helpful for debugging if the underlying error from `cmd.Run()` was included in the log or returned error string. This would provide more context on why the command failed (e.g., `gcloud` not found, not authenticated).

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+	} else {
+		log.Println(".gemini directory does not exist in /workspaces")
+	}
+
+	geminiTokenFile := filepath.Join(tokensDir, "gemini")
+	geminiKey, err := os.ReadFile(geminiTokenFile)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %v", geminiTokenFile, err)
+	}
+	os.Setenv("GEMINI_API_KEY", string(geminiKey))
```
barney-s: The `Setup` function writes the `GCP_API_KEY` to a file in the user's configuration directory. Storing secrets in plaintext on the filesystem can be a security risk. Consider whether this is the most secure way to handle the API key. If possible, using the key directly from the environment variable without writing it to a file would be preferable.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		log.Printf("gemini command failed: %v. Output: %s", err, string(output))
+		return nil, err
+	}
+
+	return stripYAMLMarkers(output), nil
+}
+
+// stripYAMLMarkers looks for ```yaml and ``` markers in the input byte slice.
+// If found, it strips these markers and returns the content between them.
```
barney-s: The prompt is passed directly as a string argument to the shell command. This could be vulnerable to shell injection if the `prompt` variable contains malicious input. It would be safer to pass the prompt to the command via standard input (stdin) to avoid shell interpretation of the prompt content.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+// If markers are not found, the original byte slice is returned.
+func stripYAMLMarkers(input []byte) []byte {
+	startMarker := []byte("```yaml")
```
barney-s: The error message "failed to run gemini cli" is generic. When `cmd.Run()` fails, it would be very helpful to capture and include the `stderr` from the command in the returned error. This would give the caller immediate insight into why the execution failed.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %v", geminiTokenFile, err)
+	}
+	os.Setenv("GEMINI_API_KEY", string(geminiKey))
+	return nil
+}
+
+func (g *Gemini) Run(agentPrompt string) ([]byte, error) {
+	log.Println("running gemini")
+
```
barney-s: The file permissions for the credentials file are set to `0644`, which makes the file world-readable. Since this file contains a sensitive API key, the permissions should be more restrictive, such as `0600`, to ensure only the file owner can read it.

repo-agent/llm/gemini.go
```diff
@@ -0,0 +1,104 @@
...
+func stripYAMLMarkers(input []byte) []byte {
+	startMarker := []byte("```yaml")
+	endMarker := []byte("```")
+
+	startIndex := bytes.Index(input, startMarker)
+	if startIndex == -1 {
+		return input // Start marker not found
+	}
+
+	// Adjust startIndex to point after the start marker
```
barney-s: The `stripYAMLMarkers` function uses `strings.TrimPrefix` and `strings.TrimSuffix`, which will only remove a single instance of the prefix and suffix. If the LLM response could potentially include multiple or nested YAML blocks, this might not fully clean the output. While this is an edge case, for more robust parsing, using a regular expression or a more stateful parser might be considered.

repo-agent/llm/gemini_test.go
```diff
@@ -0,0 +1,234 @@
...
+			want:  []byte("foo: bar"),
+		},
+		{
+			name:  "empty input",
+			input: []byte(""),
+			want:  []byte(""),
+		},
+		{
+			name:  "only markers",
+			input: []byte("```yaml\n```"),
```
barney-s: The test `TestNew` correctly checks that a non-nil `Gemini` object is returned. It would be more thorough to also assert that the internal fields of the struct (e.g., `project`, `location`, `model`) are initialized with the expected default values.

repo-agent/llm/gemini_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := stripYAMLMarkers(tt.input); !bytes.Equal(got, tt.want) {
+				t.Errorf("stripYAMLMarkers() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
```
barney-s: In the `TestGeminiSetup` test, the mock executor is set up to return a specific project ID. The test should also include a case where the command to get the project ID fails (i.e., the mock executor returns an error) to ensure that `Setup` correctly propagates the error.

repo-agent/pkg/llm/exec.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2025 Google LLC
+//
```
seans3: Yes. I think that's a good idea. Fixed.

repo-agent/review-sandbox/main.go
seans3: Added a slice of `PostProcessor` functions to the LLM provider, with an updated interface. The `stripYAML` function is now a `PostProcessor` function. BTW, the `stripYAML` function does nothing if there is only text. I've added unit tests for that. Please let me know what you think.

repo-agent/pkg/llm/provider.go
```diff
@@ -0,0 +1,40 @@
...
+	// AddPostProcessor adds a post-processing function to the provider.
+	// These functions are applied sequentially to the LLM's raw output.
+	AddPostProcessor(p PostProcessor)
+}
+
+func NewLLMProvider(name string) (Provider, error) {
+	switch name {
+	case "gemini-cli":
+		g := &Gemini{Executor: &RealCommandExecutor{}}
+		g.AddPostProcessor(stripYAMLMarkers)
```
barney-s: `AddPostProcessor` is welcome change. But would you move the call to AddPostProcessor to `review-sandbox/main.go` instead of building it part of the gemini provider. That was users of LLM interface can choose the processors based on their use case. For review we expect yaml formatted output.

seans3: Moved adding this post processor function to `review-sandbox/main.go`. Please let me know what you think.

PR #112, opened 2025-11-12
Basic e2e test and github action

barney-s:


dev/tools/test-e2e-chainsaw
```diff
@@ -0,0 +1,82 @@
...
+from shared import utils
+
+CHAINSAW_VERSION = "v0.2.12"
+
+def install_chainsaw(repo_root):
+    """Installs chainsaw if not present or if the version is incorrect."""
+    local_bin = os.path.join(repo_root, "bin")
+    chainsaw_path = os.path.join(local_bin, "chainsaw")
+
+    if os.path.exists(chainsaw_path):
```
aaron-prindle: The logic to check the `chainsaw` version and reinstall it is a bit complex. A simpler and more robust approach in a CI context would be to remove the existing binary unconditionally and reinstall it on every run. This ensures a clean and predictable state.

repo-agent/test/e2e/install-successful/repo-agent-components-assert.yaml
```diff
@@ -0,0 +1,53 @@
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: repo-agent-system
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pr-review-api
+  namespace: repo-agent-system
```
aaron-prindle: While checking for `readyReplicas: 1` is good, the assertion could be stronger. Consider also asserting the image tag of the deployment to ensure that the correct version of the application was deployed and is being tested.

dev/tools/install-agent-sandbox
```diff
@@ -0,0 +1,62 @@
...
+        time.sleep(5)
+
+        kubectl_get_pods_command = ["kubectl", "get", "pods", "-n", "agent-sandbox-system"]
+        run_command(kubectl_get_pods_command)
+
+        print("Agent Sandbox installed successfully.")
+
+    except requests.exceptions.RequestException as e:
+        print(f"Error fetching Agent Sandbox version: {e}")
+    except subprocess.CalledProcessError as e:
```
aaron-prindle: The generic `except Exception as e:` is too broad. It can catch and hide unexpected errors. It's better to catch more specific exceptions to understand the root cause of failures better.

dev/ci/presubmits/test-e2e-repo-agent
```diff
@@ -22,22 +22,34 @@ from shared import utils
...
-    """ invokes e2e tests in kind cluster and outputs a junit report in the ARTIFACTS dir
-
-    The ARTIFACTS environment variable is set by prow.
+    """ invokes e2e tests in kind cluster
     """
     repo_root = utils.get_repo_root()
+    working_dir = repo_root + "/repo-agent"
+    print(f"using working dir: {working_dir}")
 
     result = subprocess.run([f"{repo_root}/dev/tools/create-kind-cluster", "e2e-test", "--recreate", "--kubeconfig", f"{repo_root}/bin/KUBECONFIG"])
```
aaron-prindle: The script executes a series of commands and stops at the first failure. It would be beneficial to add more logging before each step to make it easier to debug which command failed and why.

PR #115, opened 2025-11-13
feat: add export curl command button for PR reviews

### LINKED ISSUE #113: [repo-agent] support for exporting GitHub comments API call if read-only PAT used
repo-agent should either always allow for OR via detection of the Github PAT support copying to the clipboard a "one-liner" that is the GitHub curl command that one could manually run that would do the identical GitHub API call the repo-agent would do on their behalf.  Eg instead of (when read-only PAT detected )or in addition to "Create Draft Review" there could be an "Export Review Shell Command" that outputs something like:
```
curl -L \
  -X POST \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <YOUR_TOKEN>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/<OWNER>/<REPO>/pulls/<PR_NUMBER>/reviews \
  -d '{"comments":[{"path":"<FILE_PATH>","line":<LINE_NUMBER>,"side":"RIGHT","body":"<YOUR_COMMENT>"}]}'
```

Although this would still require them to have a read-write Github PAT, users would not have to share it externally and could have more control/visibility into the command being run and could run it themselves getting the value out of the tool while mitigating giving a tool a read/write Github PAT

aaron-prindle:
fixes #113 

This PR adds an "Export Curl Command" button to allow for a usage pattern where users give read-only tokens but can then copy the full curl command that repo-agent would essentially run and run it locally with their own write access PAT.  That way they don't have to trust our MT setup, etc for trust.  Goal is to make security conscious users able to get most functionality possible as well as de-risk potential issue where users don't want to use repo-agent as it requires write-access PAT

Gist showing the exported command working:
https://gist.github.com/aaron-prindle/14d41d56d342bb35600acfc734c04198

Screenshot 2025-11-12 4 03 16 PM

image

### DIFF
--- a/repo-agent/review-ui/ui/src/App.js
+++ b/repo-agent/review-ui/ui/src/App.js
@@ -242,6 +242,73 @@ function App() {
     .catch(err => console.error("Failed to submit PR review:", err));
   };
 
+  const handleExportCurl = (id, onSuccess) => {
+    let review;
+    if (reviewViewModes[id] === 'yaml') {
+      try {
+        review = yaml.load(yamlDrafts[id]);
+      } catch (e) {
+        alert('Invalid YAML. Please fix it before exporting.');
+        return;
+      }
+    } else {
+      review = drafts[id];
+    }
+
+    if (!review || (!review.review.body?.trim() && (!review.review.comments || review.review.comments.length === 0))) {
+      alert("Please leave a review comment before Exporting.");
+      return;
+    }
+
+    try {
+      const url = new URL(activeRepo.url);
+      const pathParts = url.pathname.split('/').filter(p => p);
+      if (pathParts.length < 2) {
+        alert("Invalid repo URL format");
+        return;
+      }
+      const owner = pathParts[0];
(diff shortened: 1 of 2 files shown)


barney-s:
wondering if we need an api for this ? Is there a reason this needs to be an API ? 
Can we make it part of the yaml that is returned by the PRs api ?

aaron-prindle:
> wondering if we need an api for this ? Is there a reason this needs to be an API ? Can we make it part of the yaml that is returned by the PRs api ?

Done, I've removed the API additions for this.

Initially I had thought of making this an API to make it analagous to "Create Draft Review" so that we could hook into any kind of "suggestion accepted" hooks we had (this would be the no write-access PAT signal for "suggestion accepted").  I agree that making it part of the returned YAML (although we need to account for diffs so it isn't purely static).  I've updated to PR here to remove the API elements.  We can revisit if we need a hook for "suggestion accepted"

Issue #85, opened 2025-11-09
[repo-agent] currently repo-agent PR review comment line #s are incorrect (skewed)


aaron-prindle:
Assuming repo-agent was used to generate the comments on:

- https://github.com/gke-labs/gemini-for-kubernetes-development/pull/81
- https://github.com/gke-labs/gemini-for-kubernetes-development/pull/80

currently the line #s on the GitHub comments are not correct (off by ~20) meaning that the comment is on a line different than the line the comment refers to.  Example:
- https://github.com/gke-labs/gemini-for-kubernetes-development/pull/80#discussion_r2508357995

PR #119, opened 2025-11-13
Make the review comments movable

barney-s:
Currently gemini-cli generates comments by a offset. We are looking to fix it #85 .
However we think even under best case scenarios we may want to reanchor comments to diff line. 

This allows that. 

[move-comments.webm](https://github.com/user-attachments/assets/89c92062-9c51-438c-a688-895196e26663)

### DIFF
--- a/repo-agent/review-ui/ui/src/App.js
+++ b/repo-agent/review-ui/ui/src/App.js
@@ -232,6 +232,12 @@ function App() {
         newDraft.review = { ...newDraft.review, body: value };
       } else if (field === 'comment.body' && index !== null) {
         newDraft.review.comments[index] = { ...newDraft.review.comments[index], body: value };
+      } else if (field === 'comment.path' && index !== null) {
+        newDraft.review.comments[index] = { ...newDraft.review.comments[index], path: value };
+      } else if (field === 'comment.line' && index !== null) {
+        newDraft.review.comments[index] = { ...newDraft.review.comments[index], line: value };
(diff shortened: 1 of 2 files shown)


barney-s:
Fixes #121

seans3 (approved):
/lgtm

I wish I had more context here, and I wish we could test this. I'll wait for @aaron-prindle feedback before merging.

repo-agent/review-ui/ui/src/App.js
```diff
@@ -431,6 +437,29 @@ function App() {
...
+        commentToUpdate.path = newPath;
+        commentToUpdate.line = newLine;
+        commentToUpdate.side = newSide;
+
+        const draftYaml = yaml.dump(draftToUpdate);
+        fetch(`/api/repo/${activeRepo.namespace}/${activeRepo.name}/prs/${id}/draft`, {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ draft: draftYaml })
+        }).catch(err => console.error("Failed to save draft:", err));
```
aaron-prindle: nit: currently if we fail to save the draft the UI is still updated so the draft and the UI are out of sync.  In this case it make sense to revert back to the prior state of the saved draft

barney-s: i honestly dont know how to fix this. Also how do i verify it. Lets get it in and treat it as a bug later on.

PR #125, opened 2025-11-14
Add golang and vim extensions to the code-server

barney-s:


repo-agent/k8s/configmap-devcontainer.yaml
```diff
@@ -15,7 +15,8 @@ data:
         "ghcr.io/gke-labs/gemini-for-kubernetes-development/gemini-cli:latest": {},
         "ghcr.io/coder/devcontainer-features/code-server:1": {
           "port": 13337,
-          "host": "0.0.0.0"
+          "host": "0.0.0.0",
+          "extensions": "golang.go,vscodevim.vim"
```
aaron-prindle: nit: should we make an issue to make this configurable?  not sure how focused we want repo-agent to be (k8s go projects vs any project) but in theory users might want to add any number of extension

barney-s: The user can bring their own devcontainer and use it in their repowatch. 
This is an example.

PR #135, opened 2025-11-18
Adding installation for antigravity

barney-s:


barney-s:
merging to test build. could not test locally using devcontainer cli.

PR #162, opened 2025-11-26
make targets to uninstall for working with non-ephemeral clusters

barney-s:
Make targets for non-ephemeral (gke/gke-autopilot) clusters:

* Add make target to install latest built images from GHCR (vs released)
* Add make target to uninstall latest as well as dependent services
* update dev/tools to support uninstall for kro, gateway , agent-sandbox

Bump disk space:

* Ran into this error on autopilot: "Container issue-sandbox exceeded its local ephemeral storage limit 1Gi"

Fixed Redis permissions:
* volume mounts on GKE autopilot are restrictive. What worked in kind cluster did not work in gke autopilot

### DIFF
--- a/repo-agent/k8s/redis-statefulset.yaml
+++ b/repo-agent/k8s/redis-statefulset.yaml
@@ -14,6 +14,10 @@ spec:
       labels:
         app: redis
     spec:
+      securityContext:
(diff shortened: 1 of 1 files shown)


barney-s:
Manually verified. Merging.

PR #168, opened 2025-11-30
fix: Dont use single-branch cloning for issues

barney-s:
Envbuilder uses `main` branch when cloning a single branch and the giturl is not fork/branch specific. For some repos like k8s, main doesnt exists.

Seeing this error:
Failed to clone repository: clone https://github.com/kubernetes/kubernetes.git: couldn't find remote ref refs/heads/main

### DIFF
--- a/repo-agent/issue-sandbox/issue-sandbox-rgd.yaml
+++ b/repo-agent/issue-sandbox/issue-sandbox-rgd.yaml
@@ -137,8 +137,8 @@ spec:
                       value: /
                     #- name: ENVBUILDER_DOCKERFILE_PATH
                     #  value: ${schema.spec.devcontainerDockerfile}
-                    - name: ENVBUILDER_GIT_CLONE_SINGLE_BRANCH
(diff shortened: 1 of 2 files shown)


barney-s:
> OOC, can you explain how this fixes the issue? I am not sure how disabling this ENV var fixes the `main` branch issues (should the branch to clone be specified as a param or does this just pull an identified primary branch from Github?)

ENVBUILDER does not take a branch for cloning. It takes a clone URL. 
When registering through UI or even if we create a repowatch using kubectl, we give the giturl of the upstream repo.
There is not branch that we can specifically clone. It is the main/master default branch.

But envbuilder is hardcoded to use "main"  when we set SINGLE_BRANCH to true.
If it is false it clones the whole repo (slower).

This hardcoding would not work for some repos like k/k which has "master".

So flipping this flag off so that it works for all repos.

We need to explore if we can compute a git-url that points to master or main depending on repo. Not sure.

PR #176, opened 2025-12-02
Updates artifact to understand claude LLM provider

seans3:
* Updates more artifacts to understand `claude` LLM providerr

### DIFF
--- a/repo-agent/pkg/llm/provider.go
+++ b/repo-agent/pkg/llm/provider.go
@@ -40,7 +40,9 @@ type Provider interface {
 func NewLLMProvider(name string) (Provider, error) {
 	switch name {
 	case "gemini-cli":
-		return &Gemini{Executor: &RealCommandExecutor{}}, nil
+		g := &Gemini{Executor: &RealCommandExecutor{}}
+		g.AddPostProcessor(StripYAMLMarkers)
+		return g, nil
 	case "claude":
 		c := &Claude{}
 		c.AddPostProcessor(StripYAMLMarkers)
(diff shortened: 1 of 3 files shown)


repo-agent/review-sandbox/main.go
```diff
@@ -53,6 +53,12 @@ func runReview() error {
 	agentName := os.Getenv("AGENT_NAME")
 	log.Printf("Review with AGENT_NAME: %s", agentName)
 
+	if agentName == "claude" {
```
barney-s: https://github.com/gke-labs/gemini-for-kubernetes-development/blob/4309682386bd42cdae5caa0cf4c6e9ff657fcd97/repo-agent/pkg/llm/gemini.go#L62 

Could you set this env from the secret in Setup() for claude provider ?

repo-agent/pkg/llm/provider.go
```diff
@@ -37,7 +37,9 @@ func NewLLMProvider(name string) (Provider, error) {
 	case "gemini-cli":
 		return &Gemini{Executor: &RealCommandExecutor{}}, nil
 	case "claude":
-		return &Claude{}, nil
+		c := &Claude{}
+		c.AddPostProcessor(StripYAMLMarkers)
```
barney-s: We add this in review-sandbox/main.go:89 runReview()
do we need it here as well ?

repo-agent/review-sandbox/main.go
barney-s: Iam not sure of this approach. But for now lets have it. Would prefer to use the tokens mounted in the pod. Can fix it later.

PR #177, opened 2025-12-02
feat: add per-user github oauth token usage and background token refresh

aaron-prindle:
This PR introduces full support for GitHub App token usage where now when a user logs in the token they get from Github is used by repo-agent directly (previously used GITHUB_PAT ENV var) as well as added automatic background token refreshing. It
  shifts the repo-agent from relying on a single, static GITHUB_PAT to using per-user OAuth tokens that are securely stored and
  refreshed automatically.
  
  
Also adds two login buttons which (for oauth app usage) changes the `scopes` passed (include or exclude "repos")
  
  
image

  
### Summary
   - Updates review-api to persist user tokens to k8s secrets on login.
   - Adds PersistingTokenSource to repowatch-controller to handle auto-refreshing of expired tokens and updating the secret.
   - Grants repowatch-controller RBAC permissions to update secrets.
   - Makes GITHUB_PAT optional in Makefile and installer.sh if OAuth creds are present.
   - Injects OAuth client credentials into controller deployment for refresh flows.

### Overview:
   * OAuth-Based Multi-Tenancy:
       * Modified the review-api OAuth callback to store the user's access token, refresh token, and expiry directly into a
         user-namespaced github-pat Kubernetes Secret.
       * Updated RepoWatch resources to reference this user-specific secret, ensuring all actions (reviews, comments, commits) are
         performed with the logged-in user's identity.

   * Background Token Refreshing:
       * Implemented a new PersistingTokenSource in the repowatch-controller.
       * The controller now automatically detects expired tokens and uses the stored refresh_token to obtain a new access token from
         GitHub.
       * Updated tokens are written back to the Kubernetes Secret in real-time, ensuring long-running background tasks (like polling
         for PRs) never fail due to token expiration.

   * Optional Initial PAT:
       * Refactored installer.sh and Makefile to make GITHUB_PAT optional during installation if GITHUB_CLIENT_ID and
         GITHUB_CLIENT_SECRET are provided.
       * The system now starts in a "waiting" state with a placeholder secret and automatically begins processing once the user logs
         in via the UI.

   * RBAC & Deployment Updates:
       * Expanded repowatch-controller RBAC permissions to allow updating secrets (required for saving refreshed tokens).
       * Updated deployment manifests to inject GitHub App credentials (GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET) into the controller
         environment.

  Testing:
   * Verified OAuth login flow stores tokens correctly in the user namespace.
   * Verified repowatch-controller picks up the credentials and performs actions as the user.
   * Verified installation succeeds without a GITHUB_PAT when OAuth vars are set.

### DIFF
--- a/repo-agent/review-ui/ui/src/App.js
+++ b/repo-agent/review-ui/ui/src/App.js
@@ -778,9 +778,8 @@ function App() {
           <div className="login-actions">
             {githubAuthEnabled ? (
                 <>
-                <button className="btn btn-submit" onClick={() => handleLogin('readonly')} style={{marginBottom: '10px', backgroundColor: '#2ea44f'}}>Login with GitHub (Read-Only)</button>
-                <br/>
-                <button className="btn btn-submit" onClick={() => handleLogin('readwrite')} style={{backgroundColor: '#f78166'}}>Login with GitHub (Read-Write)</button>
+                <button className="btn btn-submit" onClick={() => handleLogin('readwrite')} style={{backgroundColor: '#0366d6', marginRight: '10px'}}>Login with GitHub (Read-Write)</button>
+                <button className="btn btn-submit" onClick={() => handleLogin('readonly')} style={{backgroundColor: '#6f42c1'}}>Login with GitHub (Read-Only)</button>
                 </>
             ) : (
                 <button className="btn btn-submit" onClick={handleGuestLogin}>Continue</button>


repo-agent/k8s/repowatch-deployment.yaml
```diff
@@ -62,6 +62,19 @@ spec:
       containers:
       - name: repowatch-controller
         image: ko://repo-agent/repowatch/cmd/repowatch-controller # placeholder value, replaced by deployment scripts
```
barney-s: It's good practice to also specify a `securityContext` at the pod level to enforce non-root execution. This enhances the security posture of the pod.

We cannot do read-only root filesystem for container since envbuilder modifies it.

repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -80,9 +136,48 @@ func NewGithubClient(ctx context.Context, k8sClient client.Client, repoWatch *re
...
+		if expiryStr != "" {
+			if expiry, err := time.Parse(time.RFC3339, expiryStr); err == nil {
+				token.Expiry = expiry
+			}
+		}
+
+		// ReuseTokenSource will use the existing token if valid, or refresh it if expired.
+		// We wrap it in PersistingTokenSource to save the new token if refreshed.
+		reusingSource := oauthConf.TokenSource(ctx, token)
+		ts = &PersistingTokenSource{
```
barney-s: The `PersistingTokenSource` is initialized with the token from the secret, but there's no locking mechanism around the token refresh and persistence. If the token expires and multiple concurrent reconciles trigger a refresh, it could lead to a race condition where multiple goroutines attempt to refresh and update the secret simultaneously. Consider adding a mutex to `PersistingTokenSource` to ensure that token refresh and persistence is an atomic operation.

repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -54,6 +56,55 @@ var seededRand = rand.New(
...
+
+func (s *PersistingTokenSource) Token() (*oauth2.Token, error) {
+	t, err := s.Source.Token()
+	if err != nil {
+		return nil, err
+	}
+
+	// Persist the token if it has changed
+	go func(token *oauth2.Token) {
+		ctx := context.Background()
```
barney-s: Starting a go routine here without any mechanism to signal cancellation (e.g., from the parent context) can lead to orphaned goroutines, especially under high load or during pod termination. The background context `context.Background()` does not carry deadlines or cancellation signals. Consider using a context derived from the reconciler's context to ensure the update operation can be cancelled if the reconciliation is cancelled.

repo-agent/k8s/rbac.generated.yaml
```diff
@@ -11,6 +11,8 @@ rules:
   verbs:
   - get
   - list
+  - patch
```
barney-s: The permissions `patch` and `update` on `secrets` are quite broad. Is it possible to restrict these permissions to specific secret name ?  something like `resourceNames` field in the RBAC role.

repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -80,9 +136,48 @@ func NewGithubClient(ctx context.Context, k8sClient client.Client, repoWatch *re
...
+	clientID := os.Getenv("GITHUB_CLIENT_ID")
+	clientSecret := os.Getenv("GITHUB_CLIENT_SECRET")
+	refreshToken := string(secret.Data["refresh_token"])
+	expiryStr := string(secret.Data["expiry"])
+
+	var ts oauth2.TokenSource
+
+	if clientID != "" && clientSecret != "" && refreshToken != "" {
+		// Use refresh flow
+		oauthConf := &oauth2.Config{
```
barney-s: Is creating a new `oauth2.Config` on every reconciliation inefficient ? If this configuration is static for the lifetime of the controller, could we initialize this once when the controller starts and reuse the same config object in all reconciliations.

Does this create a separate http.Conn each time ?

repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -54,6 +56,55 @@ var seededRand = rand.New(
...
+		secret.Data["pat"] = []byte(token.AccessToken)
+		if token.RefreshToken != "" {
+			secret.Data["refresh_token"] = []byte(token.RefreshToken)
+		}
+		if !token.Expiry.IsZero() {
+			secret.Data["expiry"] = []byte(token.Expiry.Format(time.RFC3339))
+		}
+
+		if err := s.K8sClient.Update(ctx, secret); err != nil {
+			log.Log.Error(err, "failed to update secret with new token", "secret", s.SecretName)
```
barney-s: The error message "waiting for user login to populate github token" is helpful, but the reconciler will likely retry immediately, leading to a tight loop of log spam and wasted CPU cycles. It would be better to return a `ctrl.Result` with a `RequeueAfter` duration (e.g., 1 minute) to gracefully wait for the user to log in without overwhelming the controller.

barney-s (commented):
The move away from a static GITHUB_PAT is a significant improvement towards a more secure and robust multi-tenant architecture with per-user GitHub authentication and automated token refreshes.

The overall implementation looks solid. I have a few comments regarding potential race conditions and security hardening. Please see the detailed comments for specific suggestions.

repo-agent/repowatch/controllers/repowatch_controller.go
aaron-prindle: We need to refresh oauth token and config might be dynamic in future (change read-only -> read-write), leaving as is.

repo-agent/review-ui/review-api/main.go
```diff
@@ -160,8 +161,9 @@ func initOAuth() {
 	oauthConf = &oauth2.Config{
 		ClientID:     clientID,
 		ClientSecret: clientSecret,
-		Scopes:       []string{"read:user", "user:email"},
-		Endpoint:     githuboauth.Endpoint,
+		// for Github "Github App" scopes are ignored, for Github "Oauth" app scopes are used
```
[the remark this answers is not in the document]
aaron-prindle: Updated comment to be more clear

repo-agent/k8s/rbac.generated.yaml
aaron-prindle: These permission are necessary for updating the secretes w/ the refresh tokens.  The PR is just added patch and update for this reason

repo-agent/repowatch/controllers/repowatch_controller.go
aaron-prindle: Done

repo-agent/repowatch/controllers/repowatch_controller.go
aaron-prindle: Added a lock here

PR #180, opened 2025-12-03
fix: de-dupe review note and body when running agent multiple times

barney-s:


repo-agent/review-sandbox/main.go
```diff
@@ -214,6 +223,26 @@ func runReview() error {
 
 	log.Printf("Finished agent runs. Total successful runs: %d. Total comments: %d", successfulRuns, len(accumulatedAgentOutput.Review.Comments))
 
+	if successfulRuns > 1 {
+		log.Println("Deduplicating and combining agent output text from multiple runs.")
+		combinedNote, err := dedupeAndCombineText(accumulatedAgentOutput.Note)
```
aaron-prindle: nit, should DedupeAndCombineText take the existing llm provider or create its own?  (not sure how important re-use is for the llm provider)

repo-agent/review-sandbox/main.go
```diff
@@ -401,3 +430,24 @@ func startCodeServer() (*exec.Cmd, error) {
...
+	provider, err := llm.NewLLMProvider(agentName)
+	if err != nil {
+		return "", err
+	}
+
+	if err := provider.Setup("/workspaces", "/tokens"); err != nil {
+		return "", err
+	}
+
+	prompt := fmt.Sprintf("The following text contains multiple sections separated by '---'. Please deduplicate and combine them into a single coherent section of text. Return only the result.\n\n%s", text)
```
aaron-prindle: nit: can you add a comment here explaining where the sections are concatenated with `---`? IIUC it's from  the accumulatedAgentOutput (line 207) but might make sense to comment here for locality and also give semantic meaning to what is concatenated with ---

repo-agent/review-sandbox/main.go
barney-s: fixed. passing provider.

repo-agent/review-sandbox/main.go
barney-s: added a comment in 2 places

PR #181, opened 2025-12-04
Update repowatch_controller to account for multiple LLM providers

seans3:
* Updates the `sandboxName` in the `repowatch_controller` to use the `RepoWatch.Name` instead of the repository name, which accounts for multiple `RepoWatch` with the same repository (but different LLM providers).
* Updates the logic of the `repowatch-controller` to ensure reconciliation only occurs per `RepoWatch`.
* Adds unit tests for multiple `RepoWatch` objects with same repository.
  * Unit test coverage for `repowatch/controllers` package: `81.6%`


repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -481,42 +507,23 @@ func (r *RepoWatchReconciler) reconcileReviewSandboxes(ctx context.Context, repo
 			if err := r.Delete(ctx, &sandbox); err != nil {
 				log.Error(err, "unable to delete sandbox", "sandbox", sandbox.GetName())
 			}
-			totalSandboxes--
```
barney-s: should we have this ?

seans3: No. It is a mistake. I have reintroduced this code, and added unit tests to verify it works.

PR #182, opened 2025-12-04
Dont set imagepull policy to ifnotpresent

barney-s:
When updating GKE auto[pilot cluster, the UI is serving stale image.


seans3 (approved):
/lgtm
/approve

PR #187, opened 2025-12-05
Expand / commands in system prompt

barney-s:
Issue:
When using configdir and injecting custom gemini / commands, the review system prompt will not expand it. Typical cli users would type /namespace:command args in the gemini cli. When running in a script or a process we cannot use system prompt + / command.

Fix:
We will regex and expand the / commands inline in the system prompt.

1. New File: pkg/llm/command_expansion.go - Implements the parsing of /namespace:command "args" syntax and retrieval of the prompt from the corresponding .toml command file.
3. New File: pkg/llm/command_expansion_test.go - Tests for the expansion logic.
4. Modified: pkg/llm/gemini.go - Updated Gemini.Run to invoke expandCommands on the prompt before passing it to the CLI.

You can now use /namespace:command "args" within your system prompts (either in the reviewPromptTemplate in Go code or in the prompt field of your RepoWatch CRD), and they will be expanded using the definitions in .gemini/commands.

Verified manually. Additional review instructions section has the expanded / command

```
root ➜ /workspaces $ cat agent-prompt-run1.txt 
......

----------------
additional review instructions:
Please analyze the git pull request 5770 in GoogleCloudPlatform/k8s-config-connector.
Review pull request 5770.
Please also look at any checks or comments for that pull request.

## AI Pull Request Review Assistant

This section defines your role when I ask you to review a Pull Request.
You do not have the gh tool and cannot install it.

### 1. Persona and Goal

You are an expert AI assistant specializing in reviewing code, the Config Connector (kcc or cnrm) code base and providing feedback on pull requests.
Your primary goal is to provide actionable feedback for the Pull Request author.

**DO NOT** recommend whether to merge the PR.
**DO NOT** attempt to post any comments to the PR.
**DO NOT** post a review to the PR.

### 2. Information Gathering

You must analyze the PR description, comments from all contributors, and the code changes (`diff`).
Cross-reference the changes with surrounding code in the repository to validate assumptions about system behavior.

Please call the guthub MCP server to get the details of the pull request.
Please print the title of the PR after you call it.

Read @.gemini/commands/google-internal/KCC-Review.json for information about code reviews on KCC.
Read @.gemini/commands/google-internal/GO-Review.json for information about code reviews on golang code.

### 3. Analysis & Review Generation

The analysis should contain the following.
- Positive feedback on any good features of the change.
- Actionable feedback on any code which should be changed. Any such feedback should not have already been dealt with in the latest version of the code.
- All feedback should be done with a professional tone.
- You should flag if release notes are needed in the pull request description.
- You should flag if a desired milestone is missing in the pull request description.

#### Implementation Analysis
- Are there any potential bugs, logic errors, or race conditions in the implementation?
- Does the code adhere to established k8s coding conventions and best practices?

#### Backward Compatibility
- **Crucially, analyze backward compatibility implications.**
- For changes in the `apis/` directory (e.g., to exported Go types), confirm they are **backward
  compatible**.
- For changes in the `pkg/` directory, breaking changes are permissible only if **all internal call
  sites** within the kcc project have been updated accordingly. Please verify this.
- For changes in the `experiments/` direct breaking changes are permissable.

#### Testing Gaps
- How thorough is the test coverage for the new changes?
- Are there missing unit, integration, or e2e tests for critical code paths or edge cases?


### 4. Summary of review

- Briefly explain the **purpose** and **nature** of the requested code changes.
- Please give a confidence level on the correctness and completeness of the code review.

----------------

....
```

### DIFF
--- a/repo-agent/pkg/llm/command_expansion.go
+++ b/repo-agent/pkg/llm/command_expansion.go
@@ -1,0 +1,96 @@
+package llm
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// Regex for finding commands: /namespace:command "args" or /namespace:command args
+// We'll support quoted args with simple quote handling.
+var commandRegex = regexp.MustCompile(`(?m)/([\w-]+):([\w-]+)(?:\s+(".*?"|[^\s]+))?`)
+
+func expandCommands(prompt string, geminiDir string) (string, error) {
+	return replaceCommands(prompt, geminiDir)
+}
+
+func replaceCommands(text string, geminiDir string) (string, error) {
+	// We iterate through all matches and replace them.
+	// Since we might have nested commands (unlikely but possible if a command prompt contains another command?),
+	// we'll just do one pass for now.
+
+	var errs []error
+
+	// FindAllStringSubmatchIndex is better if we want to replace ranges, but ReplaceAllStringFunc is easier.
+	result := commandRegex.ReplaceAllStringFunc(text, func(match string) string {
+		submatches := commandRegex.FindStringSubmatch(match)
+		namespace := submatches[1]
+		command := submatches[2]
+		args := ""
+		if len(submatches) > 3 {
+			args = submatches[3]
+		}
+
+		// Unquote args if quoted
+		if strings.HasPrefix(args, "\"") && strings.HasSuffix(args, "\"") {
+			args = args[1 : len(args)-1]
+			// TODO: handle escaped quotes if necessary
+		}
+
+		// Look for command file
+		cmdPath := filepath.Join(geminiDir, "commands", namespace, command+".toml")
+		if _, err := os.Stat(cmdPath); os.IsNotExist(err) {
+			// Try .json? Or maybe the user didn't define it.
+			// If not found, we return the original match (maybe it's not a command).
+			return match
+		}
+
+		content, err := os.ReadFile(cmdPath)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to read command file %s: %v", cmdPath, err))
+			return match
+		}
+
+		promptTemplate, err := extractPrompt(content, filepath.Ext(cmdPath))
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to extract prompt from %s: %v", cmdPath, err))
+			return match
+		}
+
+		promptTemplate = strings.TrimSpace(promptTemplate)
+
+		// Substitute {{args}}
+		replacement := strings.ReplaceAll(promptTemplate, "{{args}}", args)
+		return replacement
+	})
+
+	if len(errs) > 0 {
+		return result, fmt.Errorf("errors expanding commands: %v", errs)
+	}
+	return result, nil
+}
+
+func extractPrompt(content []byte, ext string) (string, error) {
+	s := string(content)
+	if ext == ".toml" {
+		// Regex for prompt = """...""" (multiline)
+		// We use (?s) to let . match newlines
+		reMultiline := regexp.MustCompile(`(?s)prompt\s*=\s*"""(.*?)"""`)
+		m := reMultiline.FindStringSubmatch(s)
+		if len(m) > 1 {
+			return m[1], nil
+		}
+
+		// Regex for prompt = "..." (single line)
+		reSingle := regexp.MustCompile(`prompt\s*=\s*"(.*?)"`)
+		m2 := reSingle.FindStringSubmatch(s)
+		if len(m2) > 1 {
+			return m2[1], nil
+		}
+		return "", fmt.Errorf("prompt field not found in TOML")
+	}
+	// Add JSON support if needed
+	return "", fmt.Errorf("unsupported command file extension: %s", ext)
+}
--- a/repo-agent/pkg/llm/command_expansion_test.go
+++ b/repo-agent/pkg/llm/command_expansion_test.go
@@ -1,0 +1,79 @@
+package llm
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestExpandCommands(t *testing.T) {
+	// Setup temporary .gemini directory
+	tempDir := t.TempDir()
+	geminiDir := filepath.Join(tempDir, ".gemini")
+	commandsDir := filepath.Join(geminiDir, "commands")
+
+	// Create namespace directory
+	nsDir := filepath.Join(commandsDir, "google-internal")
+	if err := os.MkdirAll(nsDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	// Create command file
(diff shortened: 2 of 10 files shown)


repo-agent/pkg/llm/command_expansion.go
```diff
@@ -0,0 +1,96 @@
...
+			args = submatches[3]
+		}
+
+		// Unquote args if quoted
+		if strings.HasPrefix(args, "\"") && strings.HasSuffix(args, "\"") {
+			args = args[1 : len(args)-1]
+			// TODO: handle escaped quotes if necessary
+		}
+
+		// Look for command file
```
aaron-prindle: Using `strings.Replace` with `1` will only replace the first occurrence of `{{.Args}}`. If a command's TOML file legitimately contains the placeholder multiple times, only the first one will be replaced. Consider using `strings.ReplaceAll` if all occurrences should be replaced.

repo-agent/pkg/llm/command_expansion_test.go
```diff
@@ -0,0 +1,79 @@
...
+		name     string
+		prompt   string
+		expected string
+	}{
+		{
+			name:     "Basic command",
+			prompt:   `Run this: /google-internal:review-pr "123"`,
+			expected: `Run this: Reviewing PR 123.`,
+		},
+		{
```
aaron-prindle: It's good practice to `defer cleanup()` at the beginning of the test function to ensure that the temporary directory is always removed, even if the test panics.

repo-agent/pkg/llm/command_expansion.go
```diff
@@ -0,0 +1,96 @@
...
+		namespace := submatches[1]
+		command := submatches[2]
+		args := ""
+		if len(submatches) > 3 {
+			args = submatches[3]
+		}
+
+		// Unquote args if quoted
+		if strings.HasPrefix(args, "\"") && strings.HasSuffix(args, "\"") {
+			args = args[1 : len(args)-1]
```
aaron-prindle: What is the expected behavior if the `.toml` file is syntactically incorrect? `os.ReadFile` will succeed, but subsequent parsing might fail. While there's no TOML parsing here (it seems to be treated as plain text), if that ever changes, it would be a failure point. For now, it might be worth adding a comment clarifying that the file is treated as raw text.

repo-agent/pkg/llm/command_expansion.go
```diff
@@ -0,0 +1,96 @@
+package llm
```
aaron-prindle: Consider adding a package-level comment explaining the purpose of the `llm` package and briefly what this file contributes to it.

repo-agent/pkg/llm/command_expansion.go
```diff
@@ -0,0 +1,96 @@
...
+// Regex for finding commands: /namespace:command "args" or /namespace:command args
+// We'll support quoted args with simple quote handling.
+var commandRegex = regexp.MustCompile(`(?m)/([\w-]+):([\w-]+)(?:\s+(".*?"|[^\s]+))?`)
+
+func expandCommands(prompt string, geminiDir string) (string, error) {
+	return replaceCommands(prompt, geminiDir)
+}
+
+func replaceCommands(text string, geminiDir string) (string, error) {
+	// We iterate through all matches and replace them.
```
aaron-prindle: The regular expression requires arguments to be double-quoted (`"([^"]*)"`). This means commands without arguments (e.g., `/namespace:command`) will not be matched or expanded. The regex should be updated to make the arguments optional to improve flexibility.

repo-agent/pkg/llm/command_expansion.go
aaron-prindle: The function `expandCommands` returns a `(string, error)` signature, but an error is never actually returned, only logged. Errors encountered during `os.ReadFile` should be propagated to the caller instead of just being logged. This prevents silent failures where a command is missing but the prompt processing continues as if nothing is wrong.

repo-agent/pkg/llm/command_expansion.go
```diff
@@ -0,0 +1,96 @@
...
+	// Since we might have nested commands (unlikely but possible if a command prompt contains another command?),
+	// we'll just do one pass for now.
+
```
aaron-prindle: The command path is hardcoded to start from `.gemini/commands`. This assumes that the agent is always executed from the root of the workspace. This could be brittle. Consider making this path configurable or resolving it relative to a well-known location, like the user's home directory or a configuration file.

repo-agent/pkg/llm/command_expansion_test.go
```diff
@@ -0,0 +1,79 @@
...
+	geminiDir := filepath.Join(tempDir, ".gemini")
+	commandsDir := filepath.Join(geminiDir, "commands")
+
+	// Create namespace directory
+	nsDir := filepath.Join(commandsDir, "google-internal")
+	if err := os.MkdirAll(nsDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	// Create command file
```
aaron-prindle: The test suite lacks a case for a prompt that contains a command without any arguments. Since the regex requires arguments, this is a key scenario to test to ensure the system either handles it gracefully or that the behavior is explicitly defined.

repo-agent/pkg/llm/command_expansion.go
aaron-prindle: Instead of using `strings.Replace` with a count of `1`, it would be more idiomatic and robust to use Go's `text/template` package for this substitution. It's designed for this exact purpose and would handle cases like the placeholder appearing multiple times or escaping more gracefully.

repo-agent/pkg/llm/command_expansion_test.go
```diff
@@ -0,0 +1,79 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := expandCommands(tt.prompt, geminiDir)
+			if err != nil {
+				// For unknown command, maybe we don't return error but leave it?
+				// But let's see.
+				t.Logf("Got error: %v", err)
+			}
+			// Check containment or exact match.
+			// My implementation might normalize spaces.
+			if result != tt.expected {
```
aaron-prindle: The test doesn't cover the scenario where the `.toml` file for a valid command does not contain the `{{.Args}}` placeholder. An explicit test case should be added to verify that the file content is returned as-is without any replacement.

repo-agent/pkg/llm/command_expansion_test.go
aaron-prindle: The error returned by `os.RemoveAll` is not checked. While unlikely to fail in a test environment, it's a best practice to handle all errors, for instance by logging it with `t.Logf` or `t.Errorf`.

repo-agent/pkg/llm/command_expansion.go
aaron-prindle: The regex only captures arguments in double quotes. It does not support single quotes or unquoted arguments. For better usability, consider extending the regex to handle these cases, especially for simple, single-word arguments.

PR #189, opened 2025-12-06
refactor review and issue sandbox

barney-s:
* add pkg/codeserver - for starting codeserver
* add pkg/gitcli - encapsulates Git operations with descriptive function names
* add pkg/agentoutput - encapsulate sidecar functionality
* refactor `issue-sandbox/main.go`: use gitcli package
* refactor `review-sandbox` and `issue-sandbox` to use codeserver pkg
* refactor `review-sidecar` and `issue-sidecar` to use agentoutput pkg

### DIFF
--- a/repo-agent/pkg/agentoutput/watcher.go
+++ b/repo-agent/pkg/agentoutput/watcher.go
@@ -29,7 +29,9 @@ import (
 )
 
 const (
-	outputFile = "/workspaces/agent-output.txt"
+	outputFile           = "/workspaces/agent-output.txt"
+	sleepIntervalSeconds = 10
+	AgentDraftAnnotation = "agentDraft"
 )
 
 // Run starts the sidecar watcher loop.
@@ -60,7 +62,7 @@ func Run(componentName string, gvr schema.GroupVersionResource) {
 
 	var last string
 	for {
-		time.Sleep(10 * time.Second)
+		time.Sleep(sleepIntervalSeconds * time.Second)
 		fmt.Println("watching for file", outputFile)
 		_, err := os.Stat(outputFile)
 		if os.IsNotExist(err) {
@@ -86,7 +88,7 @@ func Run(componentName string, gvr schema.GroupVersionResource) {
 			obj.SetAnnotations(make(map[string]string))
 		}
 		annotations := obj.GetAnnotations()
-		annotations["agentDraft"] = string(b)
+		annotations[AgentDraftAnnotation] = string(b)
 		obj.SetAnnotations(annotations)
 
 		if _, err := dc.Resource(gvr).Namespace(namespace).Update(context.TODO(), obj, metav1.UpdateOptions{}); err != nil {
--- a/repo-agent/pkg/codeserver/codeserver.go
+++ b/repo-agent/pkg/codeserver/codeserver.go
@@ -8,6 +8,11 @@ import (
 	"strings"
 )
 
+const (
+	CodeServerPort = 13337
+	WorkspacePath  = "/workspaces"
+)
+
 var execCommand = exec.Command
 
 func Start() (*exec.Cmd, error) {
@@ -19,7 +24,7 @@ func Start() (*exec.Cmd, error) {
 	}
 	repo := parts[1]
 	codeServerPath := "/usr/bin/code-server"
-	args := []string{"--auth=none", "--bind-addr=0.0.0.0:13337", "/workspaces/" + repo}
+	args := []string{"--auth=none", fmt.Sprintf("--bind-addr=0.0.0.0:%d", CodeServerPort), WorkspacePath + "/" + repo}
 	cmd := execCommand(codeServerPath, args...)
 	cmd.Stdout = os.Stdout
 	err := cmd.Start()


repo-agent/pkg/agentoutput/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	dc, err := dynamic.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	var last string
+	for {
+		time.Sleep(10 * time.Second)
```
seans3: Nit: might be better as a constant

repo-agent/pkg/agentoutput/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+			fmt.Printf("getting %s resource: %v\n", componentName, err)
+			continue
+		}
+
+		// update the annotation[agentDraft]
+		if obj.GetAnnotations() == nil {
+			obj.SetAnnotations(make(map[string]string))
+		}
+		annotations := obj.GetAnnotations()
+		annotations["agentDraft"] = string(b)
```
seans3: Nit: should this map key string be a constant?

repo-agent/pkg/agentoutput/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+const (
+	outputFile = "/workspaces/agent-output.txt"
+)
+
+// Run starts the sidecar watcher loop.
+// componentName is used for logging (e.g., "issue", "review").
+// gvr is the GroupVersionResource to update.
+func Run(componentName string, gvr schema.GroupVersionResource) {
```
aaron-prindle: Using `os.Exit(1)` in a library function is generally discouraged. It forces the entire application to terminate. This function should return an error instead, allowing the caller to decide how to handle the failure.

repo-agent/pkg/agentoutput/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+	fmt.Printf("starting %s sidecar\n", componentName)
+	name := os.Getenv("NAME")
+	if name == "" {
+		fmt.Println("missing NAME env")
+		os.Exit(1)
+	}
+	namespace := os.Getenv("NAMESPACE")
+	if namespace == "" {
+		fmt.Println("missing NAMESPACE env")
+		os.Exit(1)
```
aaron-prindle: Calling `panic` is not idiomatic for error handling in Go applications. It should be reserved for truly unrecoverable situations. It's better to return the error and let the calling function handle it.

repo-agent/pkg/agentoutput/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err.Error())
```
aaron-prindle: Calling `panic` is not idiomatic for error handling in Go applications. It should be reserved for truly unrecoverable situations. It's better to return the error and let the calling function handle it.

repo-agent/pkg/agentoutput/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	dc, err := dynamic.NewForConfig(config)
+	if err != nil {
```
aaron-prindle: An infinite loop with a static `time.Sleep` can be inflexible. Consider using a `time.Ticker` for a more regular and manageable interval. Additionally, this loop will never exit, which might not be desirable if the parent context is cancelled. The function should probably accept a `context.Context` and select on its `Done()` channel.

repo-agent/pkg/codeserver/codeserver.go
```diff
@@ -0,0 +1,31 @@
...
+	"os/exec"
+	"strings"
+)
+
+var execCommand = exec.Command
+
+func Start() (*exec.Cmd, error) {
+	log.Println("starting code-server")
+	repoURL := os.Getenv("GIT_HTML_URL")
+	parts := strings.Split(strings.TrimPrefix(repoURL, "https://github.com/"), "/")
```
aaron-prindle: Parsing the repository name from a URL with `strings.Split` is brittle. This will fail for URLs that don't match the exact `https://github.com/a/b/c/d` format. Consider using the `net/url` package for more robust URL parsing.

repo-agent/pkg/codeserver/codeserver.go
```diff
@@ -0,0 +1,31 @@
...
+	if len(parts) < 4 {
+		return nil, fmt.Errorf("invalid GIT_HTML_URL: %s", repoURL)
+	}
+	repo := parts[1]
+	codeServerPath := "/usr/bin/code-server"
```
aaron-prindle: Hardcoding the path `/workspaces/` makes this function less portable. This path should be passed in as an argument or read from a configuration.

repo-agent/pkg/codeserver/codeserver.go
```diff
@@ -0,0 +1,31 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+var execCommand = exec.Command
+
+func Start() (*exec.Cmd, error) {
+	log.Println("starting code-server")
```
aaron-prindle: Reading the `GIT_HTML_URL` directly from an environment variable makes this package less reusable as it creates a hidden dependency. It's better to pass this value as a parameter to the `Start` function.

repo-agent/pkg/agentoutput/watcher.go
barney-s: done

repo-agent/pkg/agentoutput/watcher.go
barney-s: done_

PR #190, opened 2025-12-08
Updates to plumb and complete multiple LLM providers

seans3:
* Refactored LLM providers (Gemini and Claude) to dynamically load API keys from mounted secrets via RGDs.
* Refactored `claude` LLM provider to use API key from mounted secret instead of using k8s client to view secret (just like the `gemini-cli` LLM provider).

### DIFF
--- a/repo-agent/repowatch/controllers/repowatch_controller.go
+++ b/repo-agent/repowatch/controllers/repowatch_controller.go
@@ -49,7 +49,6 @@ import (
 
 // Character set for the random string
 const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
-const RepoAgentSystemNamespace = "repo-agent-system"
 
 // We create a new *rand.Rand instance seeded with the current time.
 // This is crucial to get different results on each program execution.
@@ -112,7 +111,7 @@ func (s *PersistingTokenSource) Token() (*oauth2.Token, error) {
 func NewGithubClient(ctx context.Context, k8sClient client.Client, repoWatch *reviewv1alpha1.RepoWatch) (*github.Client, map[string]string, error) {
 	secret := &corev1.Secret{}
 	secretName := repoWatch.Spec.GithubSecretName
-	if err := k8sClient.Get(ctx, types.NamespacedName{Name: secretName, Namespace: RepoAgentSystemNamespace}, secret); err != nil {
+	if err := k8sClient.Get(ctx, types.NamespacedName{Name: secretName, Namespace: repoWatch.Namespace}, secret); err != nil {
 		return nil, nil, err
 	}
 	githubConfig := map[string]string{
--- a/repo-agent/repowatch/controllers/repowatch_controller_test.go
+++ b/repo-agent/repowatch/controllers/repowatch_controller_test.go
@@ -1176,7 +1176,7 @@ func TestNewGithubClient(t *testing.T) {
 			secret: &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "github-secret",
-					Namespace: RepoAgentSystemNamespace,
+					Namespace: "default",
 				},
 				Data: map[string][]byte{
 					"pat":   []byte("test-pat"),
@@ -1194,7 +1194,7 @@ func TestNewGithubClient(t *testing.T) {
 			secret: &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "github-secret-not-found",
-					Namespace: RepoAgentSystemNamespace,
+					Namespace: "default",
 				},
 				Data: map[string][]byte{
 					"pat": []byte("test-pat"),
@@ -1207,7 +1207,7 @@ func TestNewGithubClient(t *testing.T) {
 			secret: &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "github-secret",
-					Namespace: RepoAgentSystemNamespace,
+					Namespace: "default",
 				},
 				Data: map[string][]byte{},
 			},
@@ -1218,7 +1218,7 @@ func TestNewGithubClient(t *testing.T) {
 			secret: &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "github-secret",
-					Namespace: RepoAgentSystemNamespace,
+					Namespace: "default",
 				},
 				Data: map[string][]byte{
 					"pat": []byte("test-pat"),
@@ -1245,7 +1245,7 @@ func TestNewGithubClient(t *testing.T) {
 			repoWatch := &reviewv1alpha1.RepoWatch{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "test-repowatch",
-					Namespace: RepoAgentSystemNamespace,
+					Namespace: "default",
 				},
 				Spec: reviewv1alpha1.RepoWatchSpec{
 					RepoURL:          "https://github.com/test/repo",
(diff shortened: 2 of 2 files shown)


repo-agent/pkg/llm/claude.go
```diff
@@ -66,36 +56,21 @@ func (c *Claude) AddPostProcessor(p PostProcessor) {
...
 	if err != nil {
-		// If that fails, fall back to checking the environment variable
-		log.Printf("Not running in a Kubernetes cluster, checking for ANTHROPIC_API_KEY environment variable: %v", err)
+		// If reading from file fails, fall back to checking the environment variable
+		log.Printf("Failed to read API key from %s: %v", apiKeyPath, err)
 		apiKey, ok := os.LookupEnv(AnthropicAPIKeyEnvVar)
 		if !ok {
-			return fmt.Errorf("%s environment variable not set and not in a Kubernetes cluster", AnthropicAPIKeyEnvVar)
+			return fmt.Errorf("API key not found in %s or %s environment variable", apiKeyPath, AnthropicAPIKeyEnvVar)
 		}
```
aaron-prindle: While the fallback to an environment variable is good for local development, it might be surprising in a production environment. Consider adding a log line with `log.Printf` to indicate that the API key is being loaded from the environment variable. This would make debugging easier.

repo-agent/pkg/llm/claude.go
```diff
@@ -66,36 +56,21 @@ func (c *Claude) AddPostProcessor(p PostProcessor) {
...
+	apiKeyBytes, err := os.ReadFile(apiKeyPath)
 	if err != nil {
-		// If that fails, fall back to checking the environment variable
-		log.Printf("Not running in a Kubernetes cluster, checking for ANTHROPIC_API_KEY environment variable: %v", err)
+		// If reading from file fails, fall back to checking the environment variable
+		log.Printf("Failed to read API key from %s: %v", apiKeyPath, err)
 		apiKey, ok := os.LookupEnv(AnthropicAPIKeyEnvVar)
 		if !ok {
-			return fmt.Errorf("%s environment variable not set and not in a Kubernetes cluster", AnthropicAPIKeyEnvVar)
+			return fmt.Errorf("API key not found in %s or %s environment variable", apiKeyPath, AnthropicAPIKeyEnvVar)
```
aaron-prindle: It's good that you are logging the error, but it would be even better to also log the path that was tried. This will help in debugging cases where the secret is not mounted correctly.

repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -917,8 +917,9 @@ func (r *RepoWatchReconciler) createReviewSandboxForPR(ctx context.Context, repo
 					"name": repoWatch.Spec.Review.LLM.Provider,
 				},
 				"llm": map[string]interface{}{
-					"configdirRef": repoWatch.Spec.Review.LLM.ConfigdirRef,
-					"prompt":       prompt,
+					"configdirRef":     repoWatch.Spec.Review.LLM.ConfigdirRef,
+					"prompt":           prompt,
```
aaron-prindle: The key `apiKeySecretName` is used here, but in the RGD YAML files, it's `apiKeySecretName`. This is consistent, which is good. However, the `RepoWatch` API uses `APIKeySecretRef`. For consistency, consider renaming `apiKeySecretName` to `apiKeySecretRef` in the sandbox RGD and the controller.

repo-agent/repowatch/controllers/repowatch_controller.go
```diff
@@ -1003,8 +1004,9 @@ func (r *RepoWatchReconciler) createSandboxForIssueHandler(ctx context.Context,
 					"name": handler.LLM.Provider,
 				},
 				"llm": map[string]interface{}{
-					"configdirRef": handler.LLM.ConfigdirRef,
-					"prompt":       prompt,
+					"configdirRef":     handler.LLM.ConfigdirRef,
```
aaron-prindle: Similar to the comment above, consider renaming `apiKeySecretName` to `apiKeySecretRef` for consistency with the `RepoWatch` API.

repo-agent/pkg/llm/claude.go
```diff
@@ -66,36 +56,21 @@ func (c *Claude) AddPostProcessor(p PostProcessor) {
...
-
-	apiKey, ok := secret.Data[AnthropicAPIKeySecretKey]
-	if !ok {
-		return fmt.Errorf("secret %s does not contain key '%s'", AnthropicAPIKeySecretName, AnthropicAPIKeySecretKey)
-	}
-
-	c.apiKey = string(apiKey)
+	c.apiKey = string(bytes.TrimSpace(apiKeyBytes))
 	return nil
 }
```
aaron-prindle: The use of `bytes.TrimSpace` is a nice touch for robustness.

repo-agent/pkg/llm/claude_test.go
```diff
@@ -146,112 +141,90 @@ func TestClaudeRun(t *testing.T) {
...
+			setupFunc: func(tokensDir string) func() {
+				filePath := filepath.Join(tokensDir, AnthropicAPIKeySecretKey)
+				if err := os.WriteFile(filePath, []byte("  file-api-key  "), 0600); err != nil {
+					t.Fatalf("Failed to write API key file: %v", err)
+				}
+				return func() { os.Remove(filePath) }
+			},
+			expectedAPIKey: "file-api-key",
 			expectError:    false,
 		},
```
aaron-prindle: It's good that you've updated the tests to reflect the new secret loading mechanism. The use of a temporary directory for the tokens is a clean way to test this.

repo-agent/examples/claude-repowatch.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: review.gemini.google.com/v1alpha1
+kind: RepoWatch
+metadata:
+  name: claude
+spec:
+  repoURL: https://github.com/kubernetes/kubernetes
+  githubSecretName: github-pat
+  review:
+    llm:
+      provider: claude
```
aaron-prindle: It would be helpful to add a comment here explaining where the `anthropic-api-key` secret should come from and what key it should contain (e.g., `api-key`).

repo-agent/k8s/review-sandbox-rgd.yaml
```diff
@@ -15,6 +15,7 @@ spec:
       llm:
         prompt: string
         configdirRef: string | default=""
+        apiKeySecretName: string
```
aaron-prindle: Is there a reason to not have a `default` value for `apiKeySecretName`? If a user forgets to set it, the sandbox will fail to start. A default value like `""` might be safer, with the understanding that the LLM provider will then require an environment variable.

repo-agent/pkg/llm/claude.go
```diff
@@ -66,36 +56,21 @@ func (c *Claude) AddPostProcessor(p PostProcessor) {
...
+func (c *Claude) Setup(_, tokensDir string) error {
+	// Read API key from the mounted secret file
+	apiKeyPath := filepath.Join(tokensDir, AnthropicAPIKeySecretKey)
+	apiKeyBytes, err := os.ReadFile(apiKeyPath)
 	if err != nil {
-		// If that fails, fall back to checking the environment variable
-		log.Printf("Not running in a Kubernetes cluster, checking for ANTHROPIC_API_KEY environment variable: %v", err)
+		// If reading from file fails, fall back to checking the environment variable
+		log.Printf("Failed to read API key from %s: %v", apiKeyPath, err)
 		apiKey, ok := os.LookupEnv(AnthropicAPIKeyEnvVar)
```
aaron-prindle: The function signature `Setup(_, tokensDir string)` has an unused first argument. Consider removing it if it's not needed by the interface.

barney-s (approved):

PR #205, opened 2025-12-11
Make targets compatible with depoloying to gke autopilot cluster

barney-s:
- install-envoy is not needed for autopilot cluster
- use `gke-l7-regional-external-managed` class in GKE to expose to public IP
- rename GW class to gke-l7-regional-external-managed to work on kind
- TODO : dont install gw class in GKE
- Create TLS certs for https GW listener. this is required for public ip
- TODO: cert is not authenticated, required DNS


seans3 (approved):
/lgtm

We should add an explanation of the make targets (esp. these) in the README.md soon. There are a lot of make targets.  Merge at your leisure.

PR #220, opened 2025-12-17
Add HealthCheckPolicies for sandboxes in GKE clusters

### LINKED ISSUE #219: [repo-agent] Sandbox link not working on GKE autopilot installation
When we click the sandbox link to take us to the vscode, we see this error:

Image

For now we need to add a healthcheck yaml in all the sandbox RGDs conditionally.
Something like: 

```
apiVersion: networking.gke.io/v1
kind: HealthCheckPolicy
metadata:
  name: devc-gemini-for-kubernetes-development-issue-214-fix
  namespace: barney-s
spec:
  default:
    config:
      type: HTTP
      httpHealthCheck:
        # The specific path your application expects
        requestPath: /healthz
        # Using the same port defined in your Service and HTTPRoute
        port: 13337
  targetRef:
    group: ""
    kind: Service
    name: devc-gemini-for-kubernetes-development-issue-214-fix-lb
```

barney-s:
Fixes #219 

- Required to access sandbox code server
- GKE LBs by default hits / hence was failing since codeserver return 302 redirect for /
- Add HealthCheckPolicy to target /healthz instead
- This is needed only for platform=gke
- Add logic in RGD to conditionally install HealthCheckPolicy
- In Make based on platform, flip the RGD defaults
- In Make install the healthcheckpolicy CRD only for kind clusters so that RGD compiles (requires the CRDs to be present)
- make show-repo-agent-ip added
- update dev/deploy-to-kube to support --skip for conditionally skipping files

### DIFF
--- a/dev/tools/deploy-to-kube
+++ b/dev/tools/deploy-to-kube
@@ -98,6 +98,10 @@ def main(args):
                 continue
 
             path = os.path.join(root, filename)
+            rel_path = os.path.relpath(path, working_dir)
+            if args.skip_files and rel_path in args.skip_files:
+                print(f"skipping manifest {rel_path}")
+                continue
 
             with open(path, "r") as f:
                 # Use safe_load_all for multi-document YAML files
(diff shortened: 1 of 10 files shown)


barney-s:
Closing this in favor of #221

PR #221, opened 2025-12-17
re-architect: Enable secure access to sandbox

### LINKED ISSUE #219: [repo-agent] Sandbox link not working on GKE autopilot installation

barney-s:
Also fixes #219 
Obsoletes #220 

We are completely rearchitecting sandbox access. Earlier we used a HTTPRoute to route to each sandbox. But the route /sandbox/... was not protected by authentication. To make the access of sanbox secure and protected we need to rearchitect it.

### Architecture Change:

* Instead of having each Sandbox RGD create its own HTTPRoute pointing directly to the sandbox service, we now route all /sandbox/... traffic to the review-api service via the main Gateway.
* `review-api` Proxy: Add a new proxy handler in review-api that authenticates the request and then transparently forwards it to the correct internal sandbox service (devc-<name>-lb.<namespace>.svc.cluster.local).
* Cleanup: Removed the direct (unprotected) HTTPRoute definitions from the RGDs (review-sandbox, dev-sandbox, and issue-sandbox).

### Detailed Changes:

1. Modified `@k8s/ui-gateway.yaml`:
  * Updated the pr-review-api rule to match /sandbox in addition to /api. This ensures the Gateway sends sandbox traffic to your API service.

2. Updated `review-api`:
  * New File `@review-ui/review-api/pkg/api/handlers_sandbox_proxy.go`: Implemented proxySandbox, a handler that parses the namespace and name from the URL, constructs the internal service DNS name, and uses httputil.ReverseProxy to forward the request. It correctly handles path rewriting (stripping the /sandbox/ns/name prefix).
  * Modified `@review-ui/review-api/pkg/api/server.go`: Registered the new protected route group:

3. Updated RGDs:
  * Removed the httproute resource definition from:
    * @review-sandbox/review-sandbox-rgd.yaml
    * @dev-sandbox/dev-sandbox-rgd.yaml
    * @issue-sandbox/issue-sandbox-rgd.yaml

### New access

Now, when a user accesses https://<gateway>/sandbox/namespace/name, the request is:
1. Received by the Gateway.
2. Routed to review-api.
3. Authenticated by your existing AuthMiddleware.
4. Proxied to the specific sandbox service within the cluster.

### DIFF
--- a/repo-agent/review-ui/review-api/pkg/api/handlers_sandbox_proxy.go
+++ b/repo-agent/review-ui/review-api/pkg/api/handlers_sandbox_proxy.go
@@ -1,0 +1,91 @@
+package api
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+
+	"github.com/gin-gonic/gin"
+	"github.com/gke-labs/gemini-for-kubernetes-development/repo-agent/review-ui/review-api/pkg/auth"
+)
+
+func (s *Server) proxySandbox(c *gin.Context) {
+	namespace := c.Param("namespace")
+	name := c.Param("name")
+	// The path parameter will include the leading slash, e.g. "/some/file"
+	proxyPath := c.Param("path")
+
+	user := c.GetString(auth.UserKey)
+	if user == "" {
+		c.String(http.StatusUnauthorized, "Unauthorized")
+		return
+	}
+	if user != namespace {
+		c.String(http.StatusUnauthorized, "Unauthorized")
+		log.Printf("Unauthorized access %s for %s", user, c.Request.URL.String())
+		return
+	}
+
+	// The sandbox service name pattern from RGD: devc-<name>-lb
+	targetHost := fmt.Sprintf("devc-%s-lb.%s.svc.cluster.local:13338", name, namespace)
+
+	targetURL := &url.URL{
+		Scheme: "http",
+		Host:   targetHost,
+	}
+
+	proxy := httputil.NewSingleHostReverseProxy(targetURL)
+
+	//This forces the proxy to flush data to the client immediately after writing to the response body, instead of
+	// buffering. This is crucial for interactive applications and streaming protocols, ensuring low latency.
+	proxy.FlushInterval = -1
+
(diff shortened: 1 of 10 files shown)


aaron-prindle (approved):
Nice! Left one small nit related to logging, feel free to merge as is - LGTM

repo-agent/review-ui/review-api/pkg/api/handlers_sandbox_proxy.go
```diff
@@ -0,0 +1,89 @@
...
+			req.Header.Set("X-Forwarded-Proto", "https")
+		} else {
+			req.Header.Set("X-Forwarded-Proto", "http")
+		}
+	}
+
+	// Error handling
+	proxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
+		// Log the error
+		fmt.Printf("Proxy error for URL %s: %v\n", r.URL.String(), err)
```
aaron-prindle: nit: might make sense to use a structured logger here (vs fmt) as is done elsewhere

aaron-prindle:
I wonder if there a way to have a single middleware be streaming aware vs splitting across standard and sandbox groups?  I think this is fine for now but could make life easier if there was a way to do so

repo-agent/review-ui/review-api/pkg/api/handlers_sandbox_proxy.go
barney-s: switched to log.

PR #223, opened 2025-12-18
feat: add .gemini/settings.json general.previewFeatures=true entry and model=gemini-3-pro-preview to allow for Gemini 3 usage

### LINKED ISSUE #184: [repo-agent] configure gemini-cli for repo-agent to use Gemini 3
Tracks configuring gemini-cli for repo-agent to use Gemini 3.  I believe this can be done:

- using .gemini/settings.json

aaron-prindle:
fixes #184 

PR adds default .gemini/settings.json - general.previewFeatures=true,model=gemini-3-pro-preview entries to allow for Gemini 3 usage.  This adds a base .gemini/settings.json to the sandbox directory and makes it so `general.previewFeatures=true ` is also set in addition to whatever a user provides for their own .gemini/settings.json file

This also has a commit for helping with local testing to make it so /bin/keys.json is not required as IIUC it is only for deploying to GKE in a prod setting for setting up the GCS syncer

Manually tested by checking the .gemini/settings.json file in the sandbox:
image

### DIFF
--- a/repo-agent/Makefile
+++ b/repo-agent/Makefile
@@ -128,17 +128,15 @@ bin/tls.crt:
 	openssl req -x509 -nodes -days 365 -newkey rsa:2048  -keyout bin/tls.key -out bin/tls.crt  -subj "/CN=repo-agent"
 
 bin/key.json:
-	echo "Generate your own Google Cloud service account key and place it in bin/key.json"
-	echo "Follow instructions in ./syncer/README.md"
+	@echo "Warning: bin/key.json not found. This is optional for local KinD development but required for GKE/GCS sync."
+	@echo "To generate: Follow instructions in ./syncer/README.md"
 
 .PHONY: create-secrets
-create-secrets: bin/tls.crt
+create-secrets: bin/tls.crt bin/key.json
 	kubectl create namespace repo-agent-system || true
 	@kubectl create secret -n repo-agent-system tls repo-agent-tls --cert=bin/tls.crt --key=bin/tls.key --dry-run=client -o yaml | kubectl apply -f -
 	@if [ -f ./bin/key.json ]; then \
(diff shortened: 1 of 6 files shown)


repo-agent/pkg/llm/gemini.go
```diff
@@ -59,6 +60,11 @@ func (g *Gemini) Setup(workspacesDir, tokensDir string) error {
 		log.Println(".gemini directory does not exist in /workspaces")
 	}
 
+	// Ensure settings.json has previewFeatures: true
+	if err := ensureSettings(".gemini"); err != nil {
```
seans3: Minor nit: should `.gemini` be a constant? Do we use it elsewhere?

repo-agent/pkg/llm/gemini.go
```diff
@@ -68,6 +74,48 @@ func (g *Gemini) Setup(workspacesDir, tokensDir string) error {
...
+
+	general, ok := settings["general"].(map[string]interface{})
+	if !ok {
+		// handle case where "general" is not a map
+		general = make(map[string]interface{})
+		settings["general"] = general
+	}
+	general["previewFeatures"] = true
+	if _, ok := settings["model"]; !ok {
+		settings["model"] = "gemini-3-pro-preview"
```
seans3: Nit: maybe this is another constant (along with the new 3-flash?)

barney-s: I like this we are injecting opinionated settings. 
This will override user provided one via configdir

repo-agent/Makefile
```diff
@@ -132,10 +132,14 @@ bin/key.json:
 	echo "Follow instructions in ./syncer/README.md"
 
 .PHONY: create-secrets
-create-secrets: bin/tls.crt bin/key.json
```
barney-s: we can still retain this target and print the warning there.

aaron-prindle: done

PR #249, opened 2025-12-28
fix: cleanup redis cache when deleting repowatch

barney-s:
Issue:
When we delete a repo from the UI, it was leaving behind redis cache data.
So next time we add the same repo back cached data is showing in UI.

Fix: 
The DeleteRepo handler in the backend (review-ui/review-api/pkg/store/redis.go) has been updated.
Previously, it only deleted the repository metadata from Redis.
It has been modified it to also clean up all associated data, including Pull Requests, Issues, and Dev Sandboxes, by scanning and deleting keys that match the repository's pattern.
This ensures that when a repository is deleted and then re-added, no stale review data from the previous instance will persist.


barney-s:
verified on local cluster.
merging this since the preview cluster has data that needs to be cleaned up.

PR #250, opened 2025-12-28
fix: use correct devcontainer config

barney-s:
was causing pod to be stuck in init


barney-s:
merging this since it is a simple config change for an example

PR #268, opened 2026-01-08
dev-sandbox: create experimental command to respond to github feedback

justinsb:


repo-agent/pkg/github/client.go
```diff
@@ -0,0 +1,31 @@
+package github
+
+import (
```
barney-s: may be this org is better. We can coverge later. 
<service>/client.go vs clients/<service>.go

we currently have clients/<service>.go

PR #272, opened 2026-01-08
feat: error out if too many files were changed in a review

barney-s:
* default is 30
* can be changed by specifying repowatch `.spec.review.maxReviewFiles`

image

image

With custom no of files 77

image

### DIFF
--- a/repo-agent/review-sandbox/main_test.go
+++ b/repo-agent/review-sandbox/main_test.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestGetMaxReviewFiles(t *testing.T) {
+	// Test default value
+	os.Unsetenv("MAX_REVIEW_FILES")
+	if val := getMaxReviewFiles(); val != DefaultMaxReviewFiles {
+		t.Errorf("Expected default %d, got %d", DefaultMaxReviewFiles, val)
+	}
+
+	// Test valid value
+	os.Setenv("MAX_REVIEW_FILES", "50")
+	if val := getMaxReviewFiles(); val != 50 {
+		t.Errorf("Expected 50, got %d", val)
+	}
+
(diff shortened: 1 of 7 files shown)


repo-agent/review-sandbox/main_test.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestGetMaxReviewFiles(t *testing.T) {
+	// Test default value
+	os.Unsetenv("MAX_REVIEW_FILES")
```
justinsb: So I would probably implement this with the options pattern:

```
type ReviewSandboxOptions struct {
...
  MaxReviewFiles int
...
}

func (o *ReviewSandboxOptions) InitDefaults() {
  // MaxReviewFiles is the max files to review in a single PR
  o.MaxReviewFiles = 30 // No need for a const when it is used in one place, IMO
} 

func (o *ReviewSandboxOptions) SetDefaultsFromEnv() error {
  if s := os.Getenv("MAX_REVIEW_FILES"); s != "" {
    v, err := strconv.Atoi(s)
    if err != nil { ... }
    o.MaxReviewFiles = v
  }
....
}

func (o *ReviewSandboxOptions) Validate() []error {
  var errs []error
  ...
  if o.MaxReviewFiles > 640 {
    errs = append(errs, fmt.Errorf("maxReviewFiles was %d; 640 is enough for anyone")
  }
...
}

func (o *ReviewSandboxOptions) BindToFlags(flags *cobra.Flags) {
   flags.SetIntVar(&o.MaxReviewFiles, "max-review-files", o.MaxReviewFiles, "max files to review in a single PR")
}
```

Should be much easier to test - the trickiest one is probably SetDefaultsFromEnv and _probably_ we could do some flag-reflection instead to do this generically.

PR #276, opened 2026-01-09
Increase resources for repowatch controller 

barney-s:
The repowatch controller is very slow to react to changes. 
Adding concurrency and Bumping cpu and mem

### DIFF
--- a/repo-agent/k8s/repowatch-deployment.yaml
+++ b/repo-agent/k8s/repowatch-deployment.yaml
@@ -66,6 +66,8 @@ spec:
       containers:
       - name: repowatch-controller
         image: ko://repo-agent/repowatch/cmd/repowatch-controller # placeholder value, replaced by deployment scripts
+        args:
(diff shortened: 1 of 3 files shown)


aaron-prindle:
Also can you confirm that the repowatch controller has no shared state that concurrency might negatively impact (and we might need to lock, etc.).  Likely minor issue anyway w.r.t to what would happen w/ a race but wanted to understand if this was a possible concern here.

barney-s:
> Also can you confirm that the repowatch controller has no shared state that concurrency might negatively impact (and we might need to lock, etc.). Likely minor issue anyway w.r.t to what would happen w/ a race but wanted to understand if this was a possible concern here.

At this point we will have to discover it.

barney-s:
> /lgtm
> 
> Q: Are we concerned that doing this will increase Github API quota usage? Have we seen any Github related quota issues?

Not at the moment. But a valid point.

PR #277, opened 2026-01-09
dev-sandbox: more end-to-end functionality for github commands

justinsb:
This should be the complete functionality necessary for an end-to-end
  workflow.  We will want to iterate to make this more generic in future,
  but we have to start with the full journey and then generalize.


justinsb:
Builds on #268

PR #299, opened 2026-01-14
feat: support for prebuilt Docker image in dev-sandbox

barney-s:
Currently we support devcontainer via configmap in repowatch and when creating sandbox via cli/ui. This takes >10 mins in autopilot cluster. Support prebuilt docker image for dev sandboxes (say generic-go).

Can be used via:
1. set RepoWatch `.spec.dev.image`
2. dev-sandbox cli: `--image` parameter

## feat: support for prebuilt Docker image

1. Updated RepoWatch API Definition
* Added an optional Image field to DevSpec struct. This allows users to specify a custom image in the RepoWatch CR.
3. Updated Sandbox Logic (`dev_sandbox.go`):
* Updated DevSandboxOptions to include the Image field.
* Updated NewDevSandbox to map the Image option to the spec.image field of the generated DevSandbox resource.
4. Updated Controller Logic (`repowatch_controller.go`):
* Updated createDevSandbox to pass the Image from repowatch.spec.dev to the DevSandboxOptions.
5. Updated DevSandbox RGD (`dev-sandbox-rgd.yaml`):
* Added image parameter to the DevSandbox CRD definition within the RGD.
* Unified Agent Injection: Added an initContainer named inject-agent that copies the dev-sandbox binary from the official image to a shared emptyDir volume agent-bin mounted at /opt/repo-agent. This ensures the agent binary is available regardless of the image used.
   * Conditional Image Usage:
   * The dev-sandbox container now uses a CEL expression for the image field: image: ${schema.spec.image != "" ? schema.spec.image : "ko://repo-agent/images/dev-sandbox"}.
   * The command field is also conditional: command: ${schema.spec.image != "" ? ["/opt/repo-agent/dev-sandbox", "init"] : []}. If a custom image is provided, it forces the command to run the injected agent. If not (default), it passes an empty list (which omitempty should handle) to allow envbuilder to run as the default entrypoint.
   * Updated Env Vars: Updated ENVBUILDER_INIT_SCRIPT to point to the new shared location: /opt/repo-agent/dev sandbox init. This works for both the default envbuilder flow and the custom image flow (where the command explicitly runs it).

These changes ensure that:
* Users can provide a custom prebuilt image via repowatch.spec.dev.image.
* If provided, this image is used instead of the default envbuilder image.
* The dev-sandbox agent is always available and executed, ensuring the sandbox functions correctly (git operations,code-server, etc.).
* The default behavior (using envbuilder) is preserved when no image is specified.

## feat: add inject command to install binary
    
1. Created `repo-agent/pkg/commands/inject.go`: Implemented the inject command which copies the current binary to a specified path.
2. Modified `repo-agent/cmd/dev-sandbox/main.go`: Registered the inject command in the dev-sandbox binary.
3. Updated `repo-agent/cmd/dev-sandbox/dev-sandbox-rgd.yaml`: Changed the inject-agent init container to use the new command ["/repo-agent/dev-sandbox", "inject", "--path", "/opt/repo-agent"] instead of /bin/sh.
4. Updated `repo-agent/k8s/dev-sandbox-rgd.yaml`: Applied the same fix to the Kubernetes manifest for consistency.

## feat: add support for cloning source code
1. envbuilder was cloning the source code when using devcontainer.
2. Added similar functionality when using custom docker image

## feat: support image in cli and UI create paths
1. Add --image in the cli to provide custom image and use it when creating devsandbox
2. When creating a devsandbox via UI, use the custom image from repowatch `spec.dev`

### DIFF
--- a/repo-agent/cmd/dev-sandbox/main.go
+++ b/repo-agent/cmd/dev-sandbox/main.go
@@ -101,17 +101,17 @@ func InitContainer(ctx context.Context) error {
 		CloneURL:     os.Getenv("GIT_CLONE_URL"),
 		Destination:  repoDir,
 	}
-	if err := ib.CloneRepo(ctx); err != nil {
-		_ = agentoutput.SetAgentState(ctx, gvr, "error", fmt.Sprintf("cloning repo failed: %v", err))
-		return fmt.Errorf("Cloning repo failed: %w", err)
-	}
-
 	// if repoDir doesnt exist, we need to clone it
 	if _, err := os.Stat(repoDir); os.IsNotExist(err) {
-		if err := ib.InstallDotfilesRepo(ctx); err != nil {
-			// Note: we don't fail the entire startup if dotfiles installation fails
-			log.Error(err, "installing dotfiles repo", "repo", ib.DotFilesRepo)
+		if err := ib.CloneRepo(ctx); err != nil {
+			_ = agentoutput.SetAgentState(ctx, gvr, "error", fmt.Sprintf("cloning repo failed: %v", err))
+			return fmt.Errorf("Cloning repo failed: %w", err)
 		}
+	}
+
+	if err := ib.InstallDotfilesRepo(ctx); err != nil {
+		// Note: we don't fail the entire startup if dotfiles installation fails
+		log.Error(err, "installing dotfiles repo", "repo", ib.DotFilesRepo)
 	}
 
 	// Change to repo dir
--- a/repo-agent/pkg/api/handlers_dev.go
+++ b/repo-agent/pkg/api/handlers_dev.go
@@ -78,6 +78,8 @@ func (s *Server) createDevSandbox(c *gin.Context) {
 	apiKeySecretRef, _, _ := unstructured.NestedString(rw.Object, "spec", "dev", "llm", "apiKeySecretRef")
 	configdirRef, _, _ := unstructured.NestedString(rw.Object, "spec", "dev", "llm", "configdirRef")
 	llmProvider, _, _ := unstructured.NestedString(rw.Object, "spec", "dev", "llm", "provider")
+	image, _, _ := unstructured.NestedString(rw.Object, "spec", "dev", "image")
+	devContainerConfigRef, _, _ := unstructured.NestedString(rw.Object, "spec", "dev", "devcontainerConfigRef")
 
 	// Fetch user info from secret
 	var userName, userEmail string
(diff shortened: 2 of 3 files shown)


repo-agent/cmd/dev-sandbox/dev-sandbox-rgd.yaml
```diff
@@ -66,20 +67,16 @@ spec:
                   volumeMounts:
                     - name: workspaces-pvc
                       mountPath: /workspaces
+                - name: inject-agent
+                  image: ko://repo-agent/images/dev-sandbox
```
cheftako: The `inject-agent` init container explicitly uses `ko://repo-agent/images/dev-sandbox`. Verify that this image is accessible in all environments where the custom image might be used (e.g. if the custom image is from a private registry and the cluster has restricted internet access, they might need to ensure they can pull this `ko` image as well).

repo-agent/cmd/dev-sandbox/dev-sandbox-rgd.yaml
```diff
@@ -129,6 +126,8 @@ spec:
                           optional: true
                     - name: GIT_PUSH_ENABLED
                       value: ${string(schema.spec.destination.pushEnabled)}
+                    - name: GIT_CLONE_URL
```
cheftako: You've added `GIT_CLONE_URL`. This duplicates `ENVBUILDER_GIT_URL` to support the custom image path (where `envbuilder` isn't running). Just confirming that `schema.spec.source.cloneURL` contains the necessary authentication tokens if the repo is private, as `imagebuilder.CloneRepo` will rely on it.

repo-agent/cmd/dev-sandbox/dev-sandbox-rgd.yaml
```diff
@@ -155,9 +154,13 @@ spec:
                     - name: devcontainer-config
                       mountPath: /devcontainer.json
                       subPath: devcontainer.json
+                    - name: agent-bin
+                      mountPath: /opt/repo-agent
```
cheftako: `mountPath: /opt/repo-agent` is hardcoded here and in the `inject` command args. If this path ever needs to change, it will require updates in multiple places. It might be worth documenting this dependency clearly.

repo-agent/pkg/imagebuilder/imagebuilder.go
```diff
@@ -13,12 +13,32 @@ import (
 
 // ImageBuilder is responsible for initializing the container, e.g. installing dotfiles
 type ImageBuilder struct {
+	DotFilesRepo string
+	CloneURL     string
+	Destination  string
+}
+
```
cheftako: `CloneRepo` returns an error if `CloneURL` is empty. Ensure that the `GIT_CLONE_URL` environment variable is guaranteed to be set in the RGD configuration, otherwise the container might crash in a loop if it's missing.

PR #300, opened 2026-01-14
refactor: dev, issue sandbox logic to pkg/commands

barney-s:
`dev-sandbox` and `issue-sandbox` utilize the `pkg/commands` pattern.

1. Refactored Entry Points: cmd/dev-sandbox/main.go and cmd/issue-sandbox/main.go
  * simplified to initialize a root cobra.Command and delegate execution to command implementations in pkg/commands.
2. New Command Logic
  * introduced new files in pkg/commands/ (dev.go, dev_daemon.go, issue.go, issue_daemon.go)
  * contains the logic previously found in the main files.
  * includes setting up the environment, handling git operations, and managing the lifecycle of the code server and agent.
3. Update entry points in RGD

### DIFF
--- a/repo-agent/cmd/dev-sandbox/main.go
+++ b/repo-agent/cmd/dev-sandbox/main.go
@@ -32,12 +32,14 @@ func run(ctx context.Context) error {
 	rootCommand := &cobra.Command{
 		Use:   "dev-sandbox",
 		Short: "Gemini Dev Sandbox Agent",
-		// Default to running the issue daemon if no subcommand is provided
+		// Default to running the dev daemon if no subcommand is provided
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 0 {
 				return fmt.Errorf("dev-sandbox command does not take any arguments")
 			}
-			return commands.RunDevDaemon(cmd.Context())
+			daemonCmd := commands.DevDaemonCommand{}
+			daemonCmd.InitDefaults()
+			return daemonCmd.Run(cmd.Context())
 		},
 	}
 	rootCommand.SilenceUsage = true  // Usage is only printed for command syntax errors
--- a/repo-agent/pkg/commands/dev.go
+++ b/repo-agent/pkg/commands/dev.go
@@ -25,25 +25,104 @@ var (
 	}
 )
 
+type DevCommand struct {
+	WorkspaceDir     string
+	RepoURL          string
+	UserDotfilesRepo string
+	CloneURL         string
+	AgentName        string
+	AgentPrompt      string
+	BranchName       string
+	PushEnabled      bool
+	GithubUserOrigin string
+	GithubUserLogin  string
+	GithubUserEmail  string
+	GithubUserName   string
+}
+
+func (c *DevCommand) InitDefaults() {
+	if c.WorkspaceDir == "" {
+		c.WorkspaceDir = "/workspaces"
+	}
+	if c.RepoURL == "" {
+		c.RepoURL = os.Getenv("GIT_HTML_URL")
+	}
+	if c.UserDotfilesRepo == "" {
+		c.UserDotfilesRepo = os.Getenv("USER_DOTFILESREPO")
+	}
+	if c.CloneURL == "" {
+		c.CloneURL = os.Getenv("GIT_CLONE_URL")
+	}
+	if c.AgentName == "" {
+		c.AgentName = os.Getenv("AGENT_NAME")
+	}
+	if c.AgentPrompt == "" {
+		c.AgentPrompt = os.Getenv("AGENT_PROMPT")
+	}
+	if c.BranchName == "" {
+		c.BranchName = os.Getenv("DEV_BRANCH")
+	}
+	if !c.PushEnabled {
+		if val := os.Getenv("GIT_PUSH_ENABLED"); val == "true" {
+			c.PushEnabled = true
+		}
+	}
+	if c.GithubUserOrigin == "" {
+		c.GithubUserOrigin = os.Getenv("GITHUB_USER_ORIGIN")
+	}
+	if c.GithubUserLogin == "" {
+		c.GithubUserLogin = os.Getenv("GITHUB_USER_LOGIN")
+	}
+	if c.GithubUserEmail == "" {
+		c.GithubUserEmail = os.Getenv("GITHUB_USER_EMAIL")
+	}
+	if c.GithubUserName == "" {
+		c.GithubUserName = os.Getenv("GITHUB_USER_NAME")
+	}
+}
+
 func BuildDevCommand() *cobra.Command {
-	return &cobra.Command{
+	devCommand := DevCommand{}
+	cmd := &cobra.Command{
 		Use:   "dev",
 		Short: "Run the dev agent setup",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 0 {
 				return fmt.Errorf("dev command does not take any arguments")
 			}
-			return RunDev(cmd.Context())
+			devCommand.InitDefaults()
+			return devCommand.Run(cmd.Context())
 		},
 	}
+
+	cmd.Flags().StringVar(&devCommand.RepoURL, "repo-url", os.Getenv("GIT_HTML_URL"), "Git HTML URL")
+	cmd.Flags().StringVar(&devCommand.UserDotfilesRepo, "user-dotfiles-repo", os.Getenv("USER_DOTFILESREPO"), "User dotfiles repo")
+	cmd.Flags().StringVar(&devCommand.CloneURL, "clone-url", os.Getenv("GIT_CLONE_URL"), "Git clone URL")
+	cmd.Flags().StringVar(&devCommand.AgentName, "agent-name", os.Getenv("AGENT_NAME"), "Agent name")
+	cmd.Flags().StringVar(&devCommand.AgentPrompt, "agent-prompt", os.Getenv("AGENT_PROMPT"), "Agent prompt")
+	cmd.Flags().StringVar(&devCommand.BranchName, "branch-name", os.Getenv("DEV_BRANCH"), "Dev branch name")
+	cmd.Flags().BoolVar(&devCommand.PushEnabled, "push-enabled", os.Getenv("GIT_PUSH_ENABLED") == "true", "Enable git push")
+	cmd.Flags().StringVar(&devCommand.GithubUserOrigin, "github-user-origin", os.Getenv("GITHUB_USER_ORIGIN"), "Github user origin")
+	cmd.Flags().StringVar(&devCommand.GithubUserLogin, "github-user-login", os.Getenv("GITHUB_USER_LOGIN"), "Github user login")
+	cmd.Flags().StringVar(&devCommand.GithubUserEmail, "github-user-email", os.Getenv("GITHUB_USER_EMAIL"), "Github user email")
+	cmd.Flags().StringVar(&devCommand.GithubUserName, "github-user-name", os.Getenv("GITHUB_USER_NAME"), "Github user name")
+
+	return cmd
 }
 
-func RunDev(ctx context.Context) error {
+func (c *DevCommand) Run(ctx context.Context) error {
 	log := klog.FromContext(ctx)
 
 	go agentoutput.Run("dev", DevGVR)
 
-	repoURL := os.Getenv("GIT_HTML_URL")
+	updateState := func(state, message string) {
+		err := agentoutput.SetAgentState(ctx, DevGVR, state, message)
+		if err != nil {
+			log.Error(err, "updating agent state failed")
+		}
+	}
+
+	repoURL := c.RepoURL
 	if repoURL == "" {
 		return fmt.Errorf("GIT_HTML_URL environment variable not set")
 	}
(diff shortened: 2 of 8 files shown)


repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+	}
+}
+
+func RunDev(ctx context.Context) error {
+	log := klog.FromContext(ctx)
+
+	go agentoutput.Run("dev", DevGVR)
+
+	repoURL := os.Getenv("GIT_HTML_URL")
```
cheftako: Directly accessing `os.Getenv` here makes this function hard to unit test and tightly coupled to the environment. Pass these values via an Options struct or function arguments.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,111 @@
...
+	if repoURL == "" {
+		return fmt.Errorf("GIT_HTML_URL environment variable not set")
+	}
+
+	parts := strings.Split(strings.TrimPrefix(repoURL, "https://github.com/"), "/")
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid GIT_HTML_URL format: %s", repoURL)
+	}
+	repoDir := filepath.Join("/workspaces", parts[1])
```
cheftako: Hardcoded path `/workspaces`. Consider making this configurable via the Options struct.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Change to repo dir
+	if err := os.Chdir(repoDir); err != nil {
```
cheftako: Calling `os.Chdir` in a library function (`pkg/commands`) changes the global state of the process and can lead to unexpected side effects for other callers. Consider passing the working directory to the functions that need it (e.g. `imagebuilder`, `sandbox` functions) instead of changing the process working directory.

repo-agent/pkg/commands/dev_daemon.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gke-labs/gemini-for-kubernetes-development/repo-agent/pkg/agentoutput"
+	"github.com/spf13/cobra"
+)
+
+func BuildDevDaemonCommand() *cobra.Command {
```
cheftako: Missing `DevDaemonOptions` struct pattern.

repo-agent/pkg/commands/dev_daemon.go
```diff
@@ -0,0 +1,42 @@
...
+func RunDevDaemon(ctx context.Context) error {
+	cscmd := CodeServerCommand{}
+
+	if err := cscmd.Start(ctx); err != nil {
+		_ = agentoutput.SetAgentState(ctx, DevGVR, "error", err.Error())
+		return fmt.Errorf("failed to start code-server: %w", err)
+	}
+
+	defer func() {
+		_ = cscmd.StopCodeServer(ctx)
```
cheftako: Error from `cscmd.StopCodeServer(ctx)` is ignored. Log this error to ensure visibility of shutdown issues.

repo-agent/pkg/commands/issue.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// Prepare git branch
+	oldCommitID, err := sandbox.PrepareGitBranch(cfg)
+	if err != nil {
+		_ = agentoutput.SetAgentState(ctx, IssueGVR, "error", err.Error())
+		return fmt.Errorf("failed to prepare git branch: %w", err)
+	}
+
+	if _, err := os.Stat("../agent-prompt.txt"); os.IsNotExist(err) {
```
cheftako: Hardcoded relative path `../agent-prompt.txt`. This relies on the current working directory being set to a specific location relative to this file, which is brittle. Make this path configurable or absolute based on the repo root.

repo-agent/cmd/dev-sandbox/main.go
```diff
@@ -48,21 +29,22 @@ func main() {
...
-	rootCommand := &cobra.Command{}
-	rootCommand.SilenceUsage = true  // Usage is only printed for command syntax errors
-	rootCommand.SilenceErrors = true // We print errors ourselves
-
-	initCommand := &cobra.Command{
-		Use: "init",
+	rootCommand := &cobra.Command{
+		Use:   "dev-sandbox",
+		Short: "Gemini Dev Sandbox Agent",
+		// Default to running the issue daemon if no subcommand is provided
```
aaron-prindle: The comment mentions "issue daemon" but this is the `dev-sandbox` command. Please update to "dev daemon".

repo-agent/pkg/commands/dev_daemon.go
```diff
@@ -0,0 +1,42 @@
...
+			return RunDevDaemon(cmd.Context())
+		},
+	}
+}
+
+func RunDevDaemon(ctx context.Context) error {
+	cscmd := CodeServerCommand{}
+
+	if err := cscmd.Start(ctx); err != nil {
+		_ = agentoutput.SetAgentState(ctx, DevGVR, "error", err.Error())
```
aaron-prindle: `agentoutput.SetAgentState` is called here, but `agentoutput.Run` (which presumably initializes the status reporter) is called inside `RunDev` (line 37).
If `cscmd.Start` fails, the error state might not be successfully reported because the agent output loop hasn't started yet.
Consider moving `agentoutput.Run` to `RunDevDaemon` or ensuring it's initialized before this call.

repo-agent/pkg/commands/issue_daemon.go
```diff
@@ -0,0 +1,39 @@
...
+			return RunIssueDaemon(cmd.Context())
+		},
+	}
+}
+
+func RunIssueDaemon(ctx context.Context) error {
+	cscmd := CodeServerCommand{}
+
+	if err := cscmd.Start(ctx); err != nil {
+		_ = agentoutput.SetAgentState(ctx, IssueGVR, "error", err.Error())
```
aaron-prindle: Same issue here: `agentoutput.SetAgentState` is called before `RunIssue` (which calls `agentoutput.Run`).

repo-agent/pkg/commands/issue_daemon.go
```diff
@@ -0,0 +1,39 @@
...
+		return fmt.Errorf("failed to start code-server: %w", err)
+	}
+
+	defer func() {
+		_ = cscmd.StopCodeServer(ctx)
+	}()
+
+	_ = RunIssue(ctx)
```
aaron-prindle: The error from `RunIssue(ctx)` is ignored (`_ = ...`).
While this might be intentional to keep the `code-server` running even if the agent fails, it would be better to log this error to stderr or the agent output so the user knows *why* the agent didn't run.
In the previous implementation, `RunAgent` errors caused the program to exit (except for Quota errors).

repo-agent/pkg/commands/dev_daemon.go
```diff
@@ -0,0 +1,42 @@
...
+		return fmt.Errorf("failed to start code-server: %w", err)
+	}
+
+	defer func() {
+		_ = cscmd.StopCodeServer(ctx)
+	}()
+
+	// Run setup and agent
+	if err := RunDev(ctx); err != nil {
```
aaron-prindle: In contrast to `issue_daemon.go`, `RunDevDaemon` returns the error from `RunDev(ctx)`.
This will likely cause the pod to crash/restart. If the goal is to allow the user to debug the environment via code-server even if the agent fails (like in the issue sandbox), consider swallowing the error here too (after logging/reporting it).
If the difference is intentional, it's worth adding a comment explaining why `dev` crashes on error while `issue` stays running.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,119 @@
...
+	repoURL := os.Getenv("GIT_HTML_URL")
+	if repoURL == "" {
+		return fmt.Errorf("GIT_HTML_URL environment variable not set")
+	}
+
+	parts := strings.Split(strings.TrimPrefix(repoURL, "https://github.com/"), "/")
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid GIT_HTML_URL format: %s", repoURL)
+	}
+	repoDir := filepath.Join("/workspaces", parts[1])
```
aaron-prindle: The path `/workspaces` is hardcoded. It would be safer to use a constant or configuration variable, or ensure this aligns with the devcontainer standard used across the project.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,119 @@
...
+	ib := imagebuilder.ImageBuilder{
+		DotFilesRepo: os.Getenv("USER_DOTFILESREPO"),
+		CloneURL:     os.Getenv("GIT_CLONE_URL"),
+		Destination:  repoDir,
+	}
+	// if repoDir doesnt exist, we need to clone it
+	if _, err := os.Stat(repoDir); os.IsNotExist(err) {
+		if err := ib.CloneRepo(ctx); err != nil {
+			_ = agentoutput.SetAgentState(ctx, gvr, "error", fmt.Sprintf("cloning repo failed: %v", err))
+			return fmt.Errorf("Cloning repo failed: %w", err)
```
aaron-prindle: Using `os.Getenv(...) == "true"` is brittle (case-sensitive). Consider using `strconv.ParseBool` to handle "True", "TRUE", "1", etc.

repo-agent/pkg/commands/issue.go
```diff
@@ -0,0 +1,90 @@
...
+	go agentoutput.Run("issue", IssueGVR)
+
+	_ = agentoutput.SetAgentState(ctx, IssueGVR, "handling issue", "")
+
+	// Create config from env vars
+	cfg := sandbox.Config{
+		AgentName:        os.Getenv("AGENT_NAME"),
+		AgentPrompt:      os.Getenv("AGENT_PROMPT"),
+		BranchName:       os.Getenv("ISSUE_BRANCH"),
+		PushEnabled:      os.Getenv("GIT_PUSH_ENABLED") == "true",
```
aaron-prindle: Same here: consider using `strconv.ParseBool` for `GIT_PUSH_ENABLED`.

repo-agent/pkg/commands/issue.go
aaron-prindle: This relies on `../agent-prompt.txt` relative to the working directory.
Unlike `RunDev`, `RunIssue` does not explicitly `os.Chdir`. Ensure the working directory is guaranteed to be correct when this runs, or use an absolute path/configuration.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+}
+
+func RunDev(ctx context.Context) error {
+	log := klog.FromContext(ctx)
+
+	go agentoutput.Run("dev", DevGVR)
+
+	repoURL := os.Getenv("GIT_HTML_URL")
```
aaron-prindle: `GIT_HTML_URL` parsing assumes a specific format (https://github.com/...). If the URL is `git@github.com:...` or has a different structure, this might panic or fail. Consider using a more robust URL parser.

repo-agent/pkg/commands/dev_daemon.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/spf13/cobra"
+)
+
+func BuildDevDaemonCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "daemon",
+		Short: "Run the dev sandbox daemon",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return fmt.Errorf("daemon command does not take any arguments")
```
aaron-prindle: Checking `len(args) != 0` is good practice, but `cobra` usually handles arguments validation if `Args` field is set (e.g., `cobra.NoArgs`). You could use that for cleaner code.

repo-agent/pkg/commands/issue_daemon.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/spf13/cobra"
+)
+
+func BuildIssueDaemonCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "daemon",
+		Short: "Run the issue agent",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return fmt.Errorf("issue command does not take any arguments")
```
aaron-prindle: Consider using `Args: cobra.NoArgs` here as well.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+)
+
+func BuildDevCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "dev",
+		Short: "Run the dev agent setup",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return fmt.Errorf("dev command does not take any arguments")
```
aaron-prindle: Consider using `Args: cobra.NoArgs`.

repo-agent/pkg/commands/issue.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+)
+
+func BuildIssueCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "issue",
+		Short: "Run the issue agent",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				return fmt.Errorf("issue command does not take any arguments")
```
aaron-prindle: Consider using `Args: cobra.NoArgs`.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,111 @@
...
+
+var (
+	DevGVR = schema.GroupVersionResource{
+		Group:    "custom.agents.x-k8s.io",
+		Version:  "v1alpha1",
+		Resource: "devsandboxes",
+	}
+)
+
+func BuildDevCommand() *cobra.Command {
```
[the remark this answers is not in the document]
barney-s: done.

repo-agent/pkg/commands/dev_daemon.go
barney-s: logging

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+			return RunDev(cmd.Context())
+		},
+	}
+}
+
+func RunDev(ctx context.Context) error {
+	log := klog.FromContext(ctx)
+
+	go agentoutput.Run("dev", DevGVR)
```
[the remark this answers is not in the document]
barney-s: we may refactor this later.

repo-agent/pkg/commands/dev.go
```diff
@@ -0,0 +1,111 @@
...
+		GithubUserOrigin: os.Getenv("GITHUB_USER_ORIGIN"),
+		GithubUserLogin:  os.Getenv("GITHUB_USER_LOGIN"),
+		GithubUserEmail:  os.Getenv("GITHUB_USER_EMAIL"),
+		GithubUserName:   os.Getenv("GITHUB_USER_NAME"),
+		ReportStatus:     false,
+	}
+
+	var b imagebuilder.ImageBuilder
+	if dotFilesRepo := os.Getenv("USER_DOTFILESREPO"); dotFilesRepo != "" {
+		_ = agentoutput.SetAgentState(ctx, DevGVR, "provisioning", "installing dotfiles")
```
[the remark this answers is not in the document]
barney-s: logging now

PR #309, opened 2026-01-15
feat: Add Image field to Reviews

barney-s:
This change enables users to specify a custom container image for PR review sandboxes via the `PRReviewSpec`. This allows for greater customization of the review environment.

Changes:
- Added `Image` field to `PRReviewSpec` in `repowatch_types.go`.
- Updated `createReviewSandboxForPR` in `repowatch_controller.go` to propagate the image setting to the created `ReviewSandbox` resource.
- Updated `review-sandbox-rgd.yaml` to support the `image` parameter and added an init container (`inject-agent`) to install the review sandbox binary, mirroring the issue sandbox pattern.
- Added `install` command (aliased to `inject`) to `review-sandbox` CLI to support the init container workflow.
- Regenerated deepcopy methods and CRDs.
- Added unit test `TestReconciler_Reconcile_PR_With_Image` to verify the functionality.

### DIFF
--- a/repo-agent/api/repowatch/v1alpha1/repowatch_types.go
+++ b/repo-agent/api/repowatch/v1alpha1/repowatch_types.go
@@ -63,6 +63,10 @@ type PRReviewSpec struct {
 	// DevcontainerConfigRef string
 	DevcontainerConfigRef string `json:"devcontainerConfigRef,omitempty"`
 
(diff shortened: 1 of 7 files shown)


barney-s:
> Q: Should we also updated any documentation or make an issue for UI elements to explain how to change this, why someone would want to, and if we have any supported alternatives.

iam working on a PR to switch all built in templates to use this.

PR #325, opened 2026-01-17
build: faster builds and tests

barney-s:
Images being skipped for kind tests:
* golang-generic: because we are pulling it from latest in ghcr.
* github-mcp-server: because we are not using it atm.

This would save us time:
- save 5m for e2e tests
- saves ~5m for local build loop

Adding a separate build target in presubmits for:
* golang-generic
* github-mcp-server


justinsb (approved):
/approve
/lgtm

PR #338, opened 2026-01-19
feat: strip "I will" thoughts from output

### LINKED ISSUE #334: [repo-agent] Seeing agent thoughts in triage issuehandler
User: barney-s

issue handler for review has agent thoughts in the text that needs to be filtered out.
We implemented filtering for review easily since it is a structured output.
Need to explore doing something similar for issue triages.

Image

barney-s:
fixes #334


aaron-prindle:
/lgtm
/approve

I am fine with this method of direct text replacement but wanted to note that likely we want to state this in our prompting ("do not say 'I will', output all final results without using I or we references") to avoid different instances of this issue (other first person usages, etc.) that the direct replacement would miss.

PR #344, opened 2026-01-20
feat: Add GitHub issue autopolling

justinsb:
We look for issues assigned to a specific user and created by an allowlist of users. For each such issue, we create a sandboxed environment to work on the issue.


repo-agent/pkg/commands/sandbox.go
```diff
@@ -0,0 +1,398 @@
...
+
+	cloneRepos := []string{
+		fmt.Sprintf("/workspaces/%s=%s", repo.FilesystemName(), repo.GitCloneURL()),
+	}
+
+	log.Info("Creating sandbox", "name", sandboxName, "repos", cloneRepos, "issue", issueURL)
+
+	container := v1.Container{}
+	container.Name = "agent"
+	container.Image = "gcr.io/justinsb-knotai-dev/generic-golang:latest"
```
aaron-prindle: Currently this uses a hardcoded image that is not public.  Might make sense to update this to a public image or have the image be built and pushed by the user.

PR #463, opened 2026-02-07
feat: inline terminal for devsandbox  and issuesandbox

### LINKED ISSUE #452: [repo-agent] Terminal access in the UI
We want to show an inline terminal in the issue and dev tabs.
Please create a separate file that hosts the terminal code.

The goal is to create a terminal in the UI that connects to a backend via a http streaming interface. 
The backend serving the http streaming interface should connect to the sandbox via kubectl exec client and start the sandbox command with sshd subcommand. 

This would create a pipeline running a ssh server on the sandbox pod.
The kubectl exec stream has to be muxed to the connection from the UI.

This is similar to how we access vscode when we click "sandbox" in the UI PRcard or DevCard.

barney-s:
Fixes #452 

Implement an inline terminal in the dev-sandbox and issue-sandbox 

## Running gemini-cli 

image

## Dev sandbox

image

## Issue sandbox 

image


barney-s:
merging since i have extensively tested this manually.

PR #557, opened 2026-02-14
setup iptables using legacy

barney-s:


barney-s:
merging since it is not used by the prod "generic-golang" image

PR #878, opened 2026-04-10
fix(repowatch): cache github user query to prevent rate limiting

barney-s:
This caches the api.github.com/user query in the repowatch reconciler to reduce the number of API calls made to GitHub and prevent 403 API rate limit exceeded errors.


cheftako (approved):
LGTM

PR #930, opened 2026-05-05
Add option to force gvisor

### LINKED ISSUE #928: It should be possible to require gvisor sandboxes
Add a command line option that will guarantee all AgentSandboxes created by repo-agent use gvisor. The option should be off by default for backwards compatibility. When the option is enabled, gvisor is ALWAYS used, no matter what is configured in RepoWatch.

mtaufen:
AI Generated. Proceed with caution.

Fixes #928


repo-agent/api/repowatch/v1alpha1/repowatch_types.go
```diff
@@ -152,6 +152,12 @@ type PRReviewSpec struct {
 	// RobotAccount to use for this handler.
 	// +kubebuilder:validation:Optional
 	RobotAccount string `json:"robotAccount,omitempty"`
+
+	// DindSupport enables docker-in-docker support for the sandbox.
+	// +kubebuilder:validation:Enum=none;gvisor;privileged
+	// +kubebuilder:default=none
+	// +kubebuilder:validation:Optional
+	DindSupport string `json:"dindSupport,omitempty"`
```
mtaufen, line 160: Hmm, what were PR review sandboxes doing before, that they didn't need this? Were they already in gvisor?

mtaufen: was it previously reusing the same setting from repowatch_types.go:DevSpec? Possible since review_sandbox.go:ReviewSandboxOptions embeds DevSandboxOptions which has the DinD setting.

mtaufen: Evidently it wasn't being set (see line 1322 where this is added for the first time for review sandboxes). So maybe this new API option makes sense.

repo-agent/pkg/sandbox/review_sandbox.go
```diff
@@ -238,6 +238,12 @@ func NewReviewSandbox(opt ReviewSandboxOptions) (*unstructured.Unstructured, *co
 					},
 					"spec": map[string]interface{}{
 						"serviceAccountName": opt.ServiceAccountName,
+						"runtimeClassName": func() interface{} {
```
mtaufen, line 241: Looks like agent_sandbox.go was already doing this, and dev_sandbox.go was already relying on the behavior in agent_sandbox.go. So just review_sandbox.go needed the update.

repo-agent/pkg/controllers/repowatch/repowatch_controller.go
```diff
@@ -256,6 +256,7 @@ type Reconciler struct {
 	NewGithubClient  githubClientFactory
 	RepoSandboxImage string
 	ConfigDirImage   string
+	ForceGvisor      bool
```
cheftako: Don't love booleans in APIs. The types file has this an enum with a privileged option. I realize flags doesn't have great support for enum but maybe consider it anyway? Alternately switch to pflag or cobra?

cheftako: Also seems good insurance against needing something like forceGvisorUltra

repo-agent/pkg/controllers/repowatch/repowatch_controller_force_gvisor_test.go
```diff
@@ -0,0 +1,261 @@
+/*
+Copyright 2025.
```
cheftako: 2026

repo-agent/pkg/controllers/repowatch/repowatch_controller_force_gvisor_test.go
```diff
@@ -0,0 +1,261 @@
...
+	resp := respFunc()
+	resp.Request = req
+	return resp, nil
+}
+
+func TestReconciler_Reconcile_ForceGvisor_AllSandboxes(t *testing.T) {
+	g := gomega.NewWithT(t)
+
+	s := runtime.NewScheme()
+	_ = clientgoscheme.AddToScheme(s)
```
cheftako: Its just a test but completely ignoring the error on the AddToScheme call seems wrong. Even though this is a test, it still seems wrong. Even if we just log the errors, it is still context you want on a failed test.

cheftako (commented):
Couple of minor issues.

repo-agent/pkg/controllers/repowatch/repowatch_controller.go
mtaufen: Do you have an API suggestion? --force-sandbox-mode="gvisor"?

repo-agent/pkg/controllers/repowatch/repowatch_controller_force_gvisor_test.go
mtaufen: And this is why I just did https://github.com/ai-on-gke/ai-factory/pull/65 on the factory!

repo-agent/pkg/controllers/repowatch/repowatch_controller_force_gvisor_test.go
mtaufen: fixed

repo-agent/pkg/controllers/repowatch/repowatch_controller.go
mtaufen: updated

cheftako (approved):
Looks good, thank you!

cheftako (approved):
Approving as per https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks

PR #976, opened 2026-05-19
refactor: move cli installer and version commands to overseer-cli admin

barney-s:


cheftako:
Looks like we need to gofmt

cheftako:
Also should add cli/cli to .gitignore

barney-s:
@cheftako fixed linter failures

cheftako:
Looks like you might also need to 'git rm cli/cli'

.gitignore
```diff
@@ -4,5 +4,5 @@ bin/
 
 # Vscode files
 .vscode
-
 .jetskicli
+cli/cli
```
cheftako, line 8: Should probably also ignore bin

barney-s:
> Looks like you might also need to 'git rm cli/cli'

It is removed.

cheftako:
> > Looks like you might also need to 'git rm cli/cli'
> 
> It is removed.

It is still showing up as a binary file when I look in "files changed" above.

Never mind. -3.95 Mb is an odd way to indicate that it is deleted.

barney-s:
it is being removed from the repo... i think it was checked in by a prev commit

cheftako (approved):
Looks good.

PR #979, opened 2026-05-19
fix: detect commit status failures and improve polling feedback in pr watch

barney-s:


factory/pkg/commands/pr.go
```diff
@@ -479,80 +479,108 @@ func runPRWatch(ctx context.Context, prURL string, interval time.Duration, dryRu
...
-						} else {
-							if err := runInvestigate(ctx, prURL, "Investigate check failures for this PR"); err != nil {
-								klog.Errorf("Investigate failed: %v", err)
-							}
-						}
+			}
+		}
+
+		if hasFailure {
+			if time.Since(lastCheckedTime) > 30*time.Minute {
```
cheftako: [NiceToHave] Should this be tunable? Somehow this seems a little off. It seems like I really care if the "failure time" is more recent than my "submit PR" time, then I want to do this.

cheftako (approved):
Net positive

factory/pkg/commands/pr.go
barney-s: this is a cool off period. 
enhanced the check to include any change in sha and in that case we trigger invesigate immediately.

PR #992, opened 2026-05-21
Add gemini-3.5-flash model support

ldanielmadariaga:
Updates all files containing gemini-3-flash-preview to also support gemini-3.5-flash.

### DIFF
--- a/factory/pkg/commands/fix.go
+++ b/factory/pkg/commands/fix.go
@@ -181,7 +181,7 @@ func runFix(ctx context.Context, targetURL, prompt, name string) error {
 		IssueComments: issueComments,
 		Instruction:   prompt,
 		Branch:        branchName,
-		Models:        []string{"gemini-3-flash-preview", "gemini-3.1-pro-preview", "gemini-2.5-pro"},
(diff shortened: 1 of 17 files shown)


factory/pkg/commands/fix.go
```diff
@@ -181,7 +181,7 @@ func runFix(ctx context.Context, targetURL, prompt, name string) error {
 		IssueComments: issueComments,
 		Instruction:   prompt,
 		Branch:        branchName,
-		Models:        []string{"gemini-3-flash-preview", "gemini-3.1-pro-preview", "gemini-2.5-pro"},
+		Models:        []string{"gemini-3.5-flash", "gemini-3-flash-preview", "gemini-3.1-pro-preview", "gemini-2.5-pro"},
```
barney-s, line 184: should it be flash and flash-preview ?

acpana: do we want to keep the old models around?

ldanielmadariaga: Per https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/gemini/3-5-flash it's just `flash`

ldanielmadariaga: We use the older models as fallbakc, based on my usage dashboard we rarely make it down the list but let's keep as is for now. Open to revising in a quarter

PR #996, opened 2026-05-21
Upgrade to gemini-3.5-flash

cheftako:
Everywhere we support falling back on quota, adding gemini-3.5-flash. It will now be the first mode we try.
We will then fall back to the existing model set we had on quota exhaustion.
There are a few places we just use 2.5.
I left those alone as I do not believe they support fallback. Also we clearly did not find it useful to use an upgraded model for these use cases.


cheftako:
Closing as https://github.com/gke-labs/gemini-for-kubernetes-development/pull/992 just merged

PR #999, opened 2026-05-21
feat(factory): introduce sandbox daemon

barney-s:
for storage initialization and periodic cleanup

This introduces 'factory daemon' as the primary entrypoint for factory sandboxes, replacing direct 'envd' execution.

Key updates:
- Storage Initialization: Automatically creates GOCACHE, GOMODCACHE, and TMPDIR on the persistent volume claim mounted at /workspaces to prevent ephemeral root disk exhaustion.
- Periodic Cleanup: Runs background tasks to periodically prune old temporary files from TMPDIR/GOTMPDIR and manage Go build cache.
- Manifest & Dockerfile Integration: Updates factory-golang Dockerfile to compile the standalone factory binary and configures AgentSandbox/ReviewSandbox manifests to execute 'factory daemon'.


cheftako (approved):
Looks good.

PR #1064, opened 2026-05-26
feat: factory gemini-cli -> antigravity-cli

barney-s:


barney-s:
Blocked.

Unlike the legacy gemini-cli, agy completely ignores GEMINI_API_KEY / ANTIGRAVITY_API_KEY environment variables and forces full Google Account OAuth sign-in.

Since these sandboxes run in a headless, automated Kubernetes cluster (with no browser and no human operator to copy/paste authorization codes at runtime), we cannot use it inside a sandbox using GEMINI keys

PR #1115, opened 2026-06-09
feat(overseer): dynamically resolve default branch and state branch inside controller container

barney-s:


barney-s:
these changes shouldn't affect any of the tests

PR #1124, opened 2026-06-12
Splitting greenfield generation into it's own workflow

ldanielmadariaga:
What the label says: Splitting workflows/kcc-example.txt into brownfield instructions and putting greenfield instructions in a different workflow file.


.agents/workflows/kcc-greenfield.txt
acpana, line 1: one meta comment that I can follow up on:

If we want the workflow to generate work for itself, I'd like the workflow to use the https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/hack/tools/greenfield/calculate_coverage.py

A "classical" tool like this to determine "what resources are missing" work better imo in this case than asking the LLM to figure it out every time.

PR #1155, opened 2026-07-01
Copy parent issue labels to created Pull Request dynamically

ldanielmadariaga:
Some chores depend on these labels so ensuring they're correctly propagated.

### DIFF
--- a/factory/pkg/commands/fix.go
+++ b/factory/pkg/commands/fix.go
@@ -246,22 +246,7 @@ func runFix(ctx context.Context, targetURL, prompt, name string, noPR, watch boo
 		issueBody = prompt
 	}
 
-	triggerLabel := "factory"
-	if cfg != nil && cfg.TriggerLabel != "" {
-		triggerLabel = cfg.TriggerLabel
-	}
-	allLabels := []string{triggerLabel}
-	if cfg != nil {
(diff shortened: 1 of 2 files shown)


factory/pkg/commands/fix_test.go
acpana, line 1: love the test files 💯

factory/pkg/commands/fix.go
```diff
@@ -387,3 +382,22 @@ func runFix(ctx context.Context, targetURL, prompt, name string, noPR, watch boo
 
 	return nil
 }
+
+func resolvePRLabels(cfg *config.FactoryConfig, issue *githubv39.Issue, isIssue bool) string {
+	triggerLabel := "factory"
```
acpana, line 387: we don't actually use the `factory` label right? I know we just moved this code around but I have only relied on `overseer` label as an actuator

acpana (approved):
one question but looks good otherwise

factory/pkg/commands/fix.go
ldanielmadariaga: I'm gonna assume this is in the factory workflow to ensure that factory and overseer don't overstep each other. 

It's part of the original code so safe to keep as is.

PR #1163, opened 2026-07-02
feat(watch): Ensure that some chore tasks are running in overseer regardless of other task limits

ldanielmadariaga:
This PR modifies the watchdog queue scheduler to bypass the maxPending limit check for general chores (task-0).

### DIFF
--- a/factory/pkg/commands/watch.go
+++ b/factory/pkg/commands/watch.go
@@ -314,6 +314,11 @@ type QueueTask struct {
 	CompletedAt time.Time `yaml:"completedAt,omitempty"`
 }
 
+type QueueTaskItem struct {
+	Filename string
+	Task     *QueueTask
+}
+
 type ChoreRunState struct {
 	LastRun time.Time `json:"lastRun"`
 }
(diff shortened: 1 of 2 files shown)


factory/pkg/commands/watch.go
```diff
@@ -1888,14 +1816,16 @@ func runWatch(ctx context.Context, owner, repo string, interval time.Duration, a
...
 
-				if activeCount >= maxPending {
+				filename := item.Filename
+				task := item.Task
+
+				isGeneralChore := task.Type == "agent-chore" && task.Number == 0
+				bypassLimit := isGeneralChore && choresSpawnedInCycle < 5
+				if activeCount >= maxPending && !bypassLimit {
 					fmt.Printf("Reached maximum pending sandboxes limit (%d). Skipping remaining queue items.\n", maxPending)
 					break
```
barney-s: Using `break` here when the `maxPending` limit is reached forces general chores to be sorted to the very front of the execution queue to ensure they bypass the limit before it stops processing.

However, this priority inversion introduces a severe task starvation risk. If there are many chores in the queue, they will consume the `maxActions` limit first, completely starving critical or urgent issue-fix/PR tasks during that scan cycle.

**Suggested Fix:**
Instead of sorting chores to the front of the queue, keep the natural priority sorting order and change this `break` to `continue`. That way:
1. Critical and urgent tasks are always evaluated first.
2. If `activeCount >= maxPending` is met, critical/urgent tasks are skipped (but we don't halt the loop).
3. General chores evaluated later in the loop can still bypass the limit and run, up to the bypass budget limit.

factory/pkg/commands/watch.go
```diff
@@ -2620,3 +2554,100 @@ func selectUserForTask(ctx context.Context, ghClient *githubv39.Client, kubeClie
...
+		"low":       6,
+	}
+	getRank := func(p string) int {
+		if r, ok := priorityRank[strings.ToLower(p)]; ok {
+			return r
+		}
+		return 5
+	}
+
+	for i := 0; i < len(tasks); i++ {
```
barney-s: The `sortQueueTasks` function dereferences `tasks[i].Task` and `tasks[j].Task` directly when accessing priorities, types, and phases (e.g., `tI.Priority`). If any `Task` pointer in the slice is `nil`, the scheduler will panic.

While `loadQueueTasks` currently initializes all `Task` pointers, future changes, external tests, or corrupted data could introduce a `nil` task. It is recommended to add a nil guard at the beginning of the inner loop:

```go
if tasks[i].Task == nil || tasks[j].Task == nil {
    continue
}
```

factory/pkg/commands/watch.go
```diff
@@ -1888,14 +1816,16 @@ func runWatch(ctx context.Context, owner, repo string, interval time.Duration, a
...
+				bypassLimit := isGeneralChore && choresSpawnedInCycle < 5
+				if activeCount >= maxPending && !bypassLimit {
 					fmt.Printf("Reached maximum pending sandboxes limit (%d). Skipping remaining queue items.\n", maxPending)
 					break
 				}
 
-				filename := item.filename
-				task := item.task
-
 				sandboxName := resolveSandboxName(ctx, kubeClient, ghClient, task.Type, task.Number, owner, repo)
```
barney-s: This line calls `resolveSandboxName` for `task.Type == "agent-chore"`, which returns a sandbox name formatted as `fix-%s-%d` or `wf-issue-%d` (see lines 138-147).

However, on timeout inside `runWatch` (around line 2045), if the task is an `agent-chore` and has an empty `SessionID`, the cleanup logic attempts to delete a sandbox formatted as `agent-%s-%d`.

This mismatch prevents the force cleanup from finding and deleting the actual sandbox on timeouts, leading to a sandbox leak. Please ensure that sandbox naming is consistent between resolution and cleanup.

factory/pkg/commands/watch.go
```diff
@@ -2620,3 +2554,100 @@ func selectUserForTask(ctx context.Context, ghClient *githubv39.Client, kubeClie
...
+			if swap {
+				tasks[i], tasks[j] = tasks[j], tasks[i]
+			}
+		}
+	}
+}
+
+func countProcessingFiles(processingDir string) int {
+	processingFiles, err := os.ReadDir(processingDir)
+	if err != nil {
```
barney-s: If `os.ReadDir(processingDir)` fails (e.g., due to permission issues), `countProcessingFiles` silently returns `0`. This masks the underlying filesystem failure and could cause the scheduler to underestimate the count of active sandboxes, potentially spawning more than `maxPending`.

Consider logging a warning using `klog.Warningf` when an error is returned:

```go
processingFiles, err := os.ReadDir(processingDir)
if err != nil {
    klog.Warningf("Failed to read processing directory %s: %v", processingDir, err)
    return 0
}
```

factory/pkg/commands/watch.go
```diff
@@ -1888,14 +1816,16 @@ func runWatch(ctx context.Context, owner, repo string, interval time.Duration, a
 				}
 				activeCount := runningCount + filesInProcessing
 
-				if activeCount >= maxPending {
+				filename := item.Filename
+				task := item.Task
+
+				isGeneralChore := task.Type == "agent-chore" && task.Number == 0
```
barney-s: Bypassing the limit for general chores is a solid improvement, but note that because all general chores (`agent-chore` with `Number: 0`) resolve to the exact same sandbox name (e.g., `fix-<repo>-0` or `wf-issue-0`), they cannot run concurrently due to the sandbox busy/running checks in `runWatch`. Therefore, they will always run sequentially within a single repository, and `choresSpawnedInCycle` will never actually reach 5 in a single watch cycle. This limit is effectively a no-op for a single repo context.

factory/pkg/commands/watch.go
```diff
@@ -1939,6 +1869,10 @@ func runWatch(ctx context.Context, owner, repo string, interval time.Duration, a
 				incomingPath := filepath.Join(incomingDir, filename)
 				processingPath := filepath.Join(processingDir, filename)
 
+				if isGeneralChore {
```
barney-s: `choresSpawnedInCycle` is incremented before the `os.Rename` operation succeeds. If `os.Rename` fails, the task is skipped via `continue`, but the counter remains incremented. This can lead to an inaccurate count of active chores processed in the cycle. It is recommended to increment `choresSpawnedInCycle` after `os.Rename` completes successfully, matching how `actionsTaken` and `filesInProcessing` are handled.

factory/pkg/commands/watch.go
```diff
@@ -2620,3 +2554,100 @@ func selectUserForTask(ctx context.Context, ghClient *githubv39.Client, kubeClie
...
+
+		tasks = append(tasks, QueueTaskItem{
+			Filename: filename,
+			Task:     &t,
+		})
+	}
+	return tasks, nil
+}
+
+func sortQueueTasks(tasks []QueueTaskItem) {
```
barney-s: The custom $O(N^2)$ sorting implementation is functional and correct, but using Go's standard library `sort.Slice` or `sort.SliceStable` is more idiomatic, highly optimized ($O(N \log N)$), and easier to maintain. Consider refactoring to use `sort.Slice` for clarity and alignment with Go's standard style.

factory/pkg/commands/watch.go
```diff
@@ -2620,3 +2554,100 @@ func selectUserForTask(ctx context.Context, ghClient *githubv39.Client, kubeClie
...
+			}
+
+			if swap {
+				tasks[i], tasks[j] = tasks[j], tasks[i]
+			}
+		}
+	}
+}
+
+func countProcessingFiles(processingDir string) int {
```
barney-s: `countProcessingFiles` suppresses errors from `os.ReadDir(processingDir)` and silently returns `0`. While this aligns with the original code, silently swallowing filesystem errors (e.g., permission issues, descriptor leaks) could complicate debugging. Consider logging the error using `klog.Errorf` before returning `0`.

PR #1169, opened 2026-07-06
Improve sync secrets validation to check that new creds are valid

ldanielmadariaga:


barney-s (approved):
thanks @ldanielmadariaga

PR #1172, opened 2026-07-06
feat(factory): skip overseer sandboxes during cleanup unless --overseer flag is passed

barney-s:
Prevent periodic and manual factory cleanup runs from unintentionally deleting active overseer sandboxes.


maqiuyujoyce (approved):
/lgtm

Issue #516, opened 2026-02-12
fr: ignore comments from "side actor"


acpana:
I see the main actor as the person who delegates `codebot-robot` (robert). When the main actor is giving feedback to robert it may be worth "gating out" any other "side actors"/ persons that did not delegate robert for said task. For instance, the GH user below actually triggers robert to completely change the approach on a PR, which the main actor may not have wanted.

<img width="843" height="934" alt="Image" src="https://github.com/user-attachments/assets/d66eaace-6483-4aa6-b269-9c681e312d65" />

barney-s:
Lets iron out the requirements before we create a PR.

I agree that responding or considering all review comments would confuse the agent OR result in subpar outcomes.
Till we come up with a long term solution for this, we should consider meaningful defaults.

Considering PR author's comments only is a safe default. But i would also want to consider the repo maintainer's comments. 
So i would consider the PR author + list of maintainer's.

We need to capture the list of maintainer's somewhere (in repowatch CRD probably).

barney-s:
we support this via #1337

PR #1173, opened 2026-07-08
 feat: Support Stop Labels & Enable Iteration on Approved PRs

barney-s:
This pull request introduces two core behavioral adjustments to `factory watch` (and Overseer) to give engineers better granular control over bot mutations while keeping bots responsive to review feedback throughout the entire PR lifecycle.

### 1. Added Support for `overseer/stop` (and `<triggerLabel>/stop`) Labels
To allow maintainers to freeze bot mutations on a specific Issue or Pull Request instantly without shutting down the entire watcher service, Overseer now recognizes stop labels across all phases of the reconciliation loop:

* **New Helper Functions (`watch.go`)**:
  * **`hasStopLabel(labels, triggerLabel)`**: Checks whether the target contains `overseer/stop` (or `<triggerLabel>/stop` if a custom `--trigger-label` is configured).
  * **`removePendingTasksForNumber(incomingDir, number)`**: Automatically finds and deletes any unstarted queue files (`*-issue-<num>.yaml` and `*-pr-<num>-*.yaml`) residing in the `incoming/` directory.

* **Multi-stage Stop Enforcement**:
  * **Queueing Issues (`queueIssueTasks`)**: Skips scanning and task creation if the issue has the stop label, instantly purging any pending `task-issue-<num>.yaml` files from the queue.
  * **Processing PRs (`processPRsFunc`)**: Skips scanning and task creation for PRs labeled with the stop label right when the PR is inspected, as well as immediately after syncing inherited labels from referenced parent issues.
  * **Auto-Discovery (`issuesCreator`)**: Prevents Overseer from automatically adding the trigger label or assigning watcher bots to open issues created by bot accounts if the issue has the stop label.
  * **Task Execution (`runRunner`)**: right before a task file transitions from `incoming/` to `processing/` to spin up a container sandbox, Overseer queries the GitHub API (`ghClient.Issues.Get`) for a live label check. If the stop label was attached after the task was queued, the task is dropped and deleted from `incoming/` before launching the sandbox container.

### 2. Removed Iteration Freeze on Approved / LGTM'd PRs
Previously, Overseer would check whether a PR had the `lgtm` or `approved` label attached and halt further automated iterations (`PR #%d is approved / LGTM'd. Ignoring new comments/feedback.`). This feature caused the bot to ignore legitimate follow-up review feedback or requests for rebasing after an approval was granted.

* **Retained Detection Code**: Kept `isApproved := isPRApprovedOrLGTM(pr, prIssue, reviews)` and added `klog.V(2).Infof(...)` debugging output so metrics and internal logging still track approved states accurately.
* **Removed Freeze Logic**: Removed the check that blocked `pr-comments` task creation (`continue`) and posted the notice (`🤖 AI Factory is ignoring new comments/feedback because this PR is already approved`).
* **Outcome**: Watcher bots now remain active and responsive to new review comments (`hasNewComments == true`) even after a PR has received an `lgtm` or `approved` label.

### DIFF
--- a/factory/pkg/commands/watch.go
+++ b/factory/pkg/commands/watch.go
@@ -535,6 +535,11 @@ func queueIssueTasks(ctx context.Context, ghClient *githubv39.Client, kubeClient
 		if cfg != nil && cfg.MinNumber > 0 && num < cfg.MinNumber {
 			continue
 		}
+		if hasStopLabel(issue.Labels, triggerLabel) {
+			klog.Infof("Skipping issue #%d because it has the stop label ('overseer/stop' or '%s/stop')", num, triggerLabel)
+			removePendingTasksForNumber(incomingDir, num)
+			continue
+		}
 		if refIssues[num] {
 			klog.Infof("Skipping issue #%d because there is already a PR referencing it.", num)
 			continue
@@ -1073,6 +1078,11 @@ func runWatch(ctx context.Context, owner, repo string, interval time.Duration, a
 				if cfg != nil && cfg.MinNumber > 0 && num < cfg.MinNumber {
 					continue
 				}
+				if hasStopLabel(prIssue.Labels, triggerLabel) {
+					klog.Infof("Skipping PR #%d because it has the stop label ('overseer/stop' or '%s/stop')", num, triggerLabel)
+					removePendingTasksForNumber(incomingDir, num)
+					continue
+				}
 				pr, _, err := ghClient.PullRequests.Get(ctx, owner, repo, num)
 				if err != nil {
 					klog.Errorf("Failed to fetch full PR #%d: %v", num, err)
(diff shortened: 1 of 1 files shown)


factory/pkg/commands/watch.go
```diff
@@ -1414,37 +1427,7 @@ func runWatch(ctx context.Context, owner, repo string, interval time.Duration, a
...
-							if !hasPostedIgnore && !dryRun {
-								addGitHubComment(ctx, ghClient, owner, repo, num, ignorePrefix+" / LGTM'd.")
-							}
-
-							state.lastCommentAddressedTime = time.Now()
-							processedPRs[num] = state
-						}
-						// Skip queueing comment task since it's approved
-						continue
+						klog.V(2).Infof("PR #%d is approved / LGTM'd", num)
```
ldanielmadariaga, line 1430: Can be a follow-up but wondering if we can do something a bit more robust here (probably at the prompt level). I've seen that the bot may just keep iterating on every small comment or piece of feedback when this is not present. 

Anyways approving it's not a big blocker but something to keep in mind.

acpana: I think your request is close to this FR

https://github.com/gke-labs/gemini-for-kubernetes-development/issues/516

Ignoring comments from side actors (in this case bots) could be a good idea! Another comment we may want to ignore are "nits" or asides (thoughts for later!)

PR #1214, opened 2026-07-22
fix(watch): paginate PR reviews and review comments when scanning for feedback

barney-s:
- Use github.ListAllReviews and github.ListAllReviewComments in PR comment scanner.
- Prevents missing inline review comments on Pull Requests with more than 30 reviews (e.g. PR #11416).


acpana (approved):
let's add a test!

PR #1248, opened 2026-07-31
feat(ui,watch): add sandbox card unpause support with idle timeout retention

barney-s:
- Record sandbox.gemini.google.com/unpaused-at timestamp when unpausing sandboxes to retain active state for at least shutdownDuration during idle checks.
- Add /unpause and /pause API routes for Overseer sandboxes and implement corresponding UI buttons on the Overseer sandbox card detail view.
- Update DevCard, IssueCard, and PrReviewCard to show Sandbox Paused status when scaled down and refine button tooltips.
- Add unit tests verifying unpaused-at annotation behavior during idle lifecycle pruning.


repo-agent/pkg/k8s/k8s.go
```diff
@@ -423,90 +346,45 @@ func (m *Manager) ScaleupSandbox(ctx context.Context, namespace, repo, prID, ann
...
 	}
+	return m.ScaledownSandboxByName(ctx, namespace, sandboxName)
+}
 
-	if annotationValue != "" {
-		metadata["annotations"] = map[string]interface{}{
-			"sandbox.gemini.google.com/prevent-auto-shutdown": annotationValue,
-		}
-	}
+func (m *Manager) ScaledownDevSandboxHelper(ctx context.Context, namespace, name string) error {
```
sdowell, line 367: What is the value of this helper function (and the Scaleup flavor below)?

factory/pkg/sandbox/sandbox.go
```diff
@@ -528,6 +533,11 @@ func SuspendIdleSandboxes(ctx context.Context, kubeClient *clients.KubernetesCli
 					lastActivity = ts
 				}
 			}
+			if tsStr, ok := annotations["sandbox.gemini.google.com/unpaused-at"]; ok {
```
sdowell, line 536: Could we move "lastActivityTimestamp" to a helper function and share with the repowatch controller?

sdowell (approved):
some nits but lgtm

factory/pkg/sandbox/sandbox.go
barney-s: naah ..we want to keep the code decoupled for now .. the UI is the only thing that is shared.

PR #1254, opened 2026-08-04
feat: add PR staleness feature

sdowell:
We have observed that overseer spends a lot cycles/tokens on stale pull requests, for example to periodically rebase. In order to minimize wasted tokens/cycles, this change introduces a staleness concept where a pull request without human action after a duration is treated as stale. A stale PR will have the overseer/stop label added, which can subsequently be removed by a human to resume overseer. This uses a default value of 1 week but can be configured on the overseer CR.

### DIFF
--- a/overseer/k8s/crds/overseer.gemini.google.com_overseers.yaml
+++ b/overseer/k8s/crds/overseer.gemini.google.com_overseers.yaml
@@ -80,6 +80,8 @@ spec:
               pollInterval:
                 default: 30m
                 type: string
+              prInactivityTimeout:
+                type: string
               repo:
                 properties:
                   issueMode:
--- a/overseer/pkg/api/v1alpha1/overseer_types.go
+++ b/overseer/pkg/api/v1alpha1/overseer_types.go
@@ -152,6 +152,10 @@ type OverseerSpec struct {
 	// +kubebuilder:default="1h"
 	SandboxIdleTimeout string `json:"sandboxIdleTimeout,omitempty"`
 
+	// PRInactivityTimeout defines the time of inactivity with no human comments before pausing automated processing on a PR (defaults to 0, which disables staleness checks).
+	// +kubebuilder:validation:Optional
+	PRInactivityTimeout *metav1.Duration `json:"prInactivityTimeout,omitempty"`
+
 	// MinNumber specifies the minimum PR/issue number to process.
 	// +kubebuilder:validation:Optional
 	MinNumber *int32 `json:"minNumber,omitempty"`
--- a/factory/pkg/commands/watch.go
+++ b/factory/pkg/commands/watch.go
@@ -33,24 +33,25 @@ import (
 )
 
 type WatchFlags struct {
-	Repo               string
-	PollInterval       time.Duration
-	Assignee           string
-	Labels             []string
-	DryRun             bool
-	WatchTimeout       time.Duration
-	MaxActions         int
-	MaxPending         int
-	Mode               string
-	QueueDir           string
-	Once               bool
-	IssueMode          string
-	PRMode             string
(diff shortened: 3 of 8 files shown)


overseer/pkg/api/v1alpha1/overseer_types.go
```diff
@@ -152,6 +152,11 @@ type OverseerSpec struct {
 	// +kubebuilder:default="1h"
 	SandboxIdleTimeout string `json:"sandboxIdleTimeout,omitempty"`
 
+	// PRInactivityTimeout defines the time of inactivity with no human comments before pausing automated processing on a PR.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="168h"
+	PRInactivityTimeout *metav1.Duration `json:"prInactivityTimeout,omitempty"`
```
sdowell, line 157: @barney-s would you prefer to use type `string` for consistency with other timeout fields? I generally prefer stricter types when possible, e.g. for apply-time validation

barney-s: this is fine. we can converge later

overseer/k8s/crds/overseer.gemini.google.com_overseers.yaml
```diff
@@ -80,6 +80,9 @@ spec:
               pollInterval:
                 default: 30m
                 type: string
+              prInactivityTimeout:
+                default: 168h
```
barney-s: does this also take days ? like 10d ?

factory/commands.test
barney-s, line 1: please remove this file from the commit

sdowell: good catch, removed

overseer/k8s/crds/overseer.gemini.google.com_overseers.yaml
sdowell: It does not - under the hood, [metav1](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/duration.go#L39) uses [time.ParseDuration](https://pkg.go.dev/time#ParseDuration), which supports hours as the largest unit

barney-s: ah i see .. maybe string then ? for consistency. i think we validate the value later on

sdowell: our parsing logic uses the same underlying [time.ParseDuration](https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/factory/pkg/commands/cleanup.go#L18) so it has the same limitation, but I don't mind switching to `string` if you prefer that. WDYT?

sdowell: I would probably lean towards using `metav1.Duration` since it's a bit more standardized and offers earlier apply-time validation, but I don't mind keeping it consistent

overseer/pkg/api/v1alpha1/overseer_types.go
```diff
@@ -152,6 +152,11 @@ type OverseerSpec struct {
 	// +kubebuilder:default="1h"
 	SandboxIdleTimeout string `json:"sandboxIdleTimeout,omitempty"`
 
+	// PRInactivityTimeout defines the time of inactivity with no human comments before pausing automated processing on a PR.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="168h"
```
barney-s: let the default be 0 ==> disabled

sdowell: sounds good, updated

barney-s (approved):
thanks @sdowell

PR #1262, opened 2026-08-05
fix: implement task queue fairness

sdowell:
Prevents task starvation in the watch runner queue caused by ordering with GitHub issue/PR creation timestamp checks.

- Add `EnqueuedAt` field to `QueueTask` set upon task queueing (falling back to file modtime or `CreatedAt` for legacy tasks).
- Implement `sortTasksFairly` to perform round-robin task selection across target entities (PRs/issues/chores) while maintaining FIFO arrival order within each entity bucket.
- Preserve priority rank and phase progression ordering across entities.
- Add unit tests in `watch_test.go` to verify fairness and anti-starvation behavior.


barney-s (approved):
thanks @sdowell

PR #1266, opened 2026-08-10
fix(tasks): add automated commitAndPush fallback to address_feedback.sh and investigate_failures.sh

barney-s:
### Summary of Changes

1. **Automated `commitAndPush` Fallback**:
   - Added `OLD_HEAD` recording in `checkoutPRBranch` across both `address_feedback.sh` and `investigate_failures.sh`.
   - Added `commitAndPush` function that runs after `runGemini`:
     - If the working directory has uncommitted/unstaged edits (`git status --porcelain` non-empty), it automatically stages (`git add .`), commits, and pushes them upstream.
     - If the agent committed or rebased locally but did not push (`OLD_HEAD != NEW_HEAD`), it pushes the commits upstream (`git push origin HEAD` or `--force`).
     - If no changes were made or committed (`git status --porcelain` empty and `OLD_HEAD == NEW_HEAD`), it exits cleanly without attempting git push (`No changes to commit.`).

2. **Parity with `iterate.sh`**:
   - Matches the battle-tested pattern in `iterate.sh` and `fix_issue.sh`, preventing changes from remaining stranded in the sandbox if the LLM finishes its turn before executing git commands.

## Root cause

Address comment task was triggered but commit was not pushed. 
we need a fallback if the model doesnt push.

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/12301#issuecomment-5236814465

image

### DIFF
--- a/factory/pkg/tasks/address_feedback.sh
+++ b/factory/pkg/tasks/address_feedback.sh
@@ -113,6 +113,7 @@ function checkoutPRBranch {
     (cd "/workspaces/${REPO_NAME}" && git merge --abort 2>/dev/null || true)
     (cd "/workspaces/${REPO_NAME}" && git cherry-pick --abort 2>/dev/null || true)
     (cd "/workspaces/${REPO_NAME}" && git reset --hard HEAD && git clean -fd && /usr/bin/gh pr checkout ${PR_NUMBER} --force && (git reset --hard @{u} 2>/dev/null || git reset --hard FETCH_HEAD))
+    OLD_HEAD=$(cd "/workspaces/${REPO_NAME}" && git rev-parse HEAD)
 }
 
 function configureGemini {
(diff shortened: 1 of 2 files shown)


factory/pkg/tasks/address_feedback.sh
```diff
@@ -317,6 +318,34 @@ function runGemini {
...
+    echo "Running commitAndPush..."
+    pushd "/workspaces/${REPO_NAME}" > /dev/null
+    
+    NEW_HEAD=$(git rev-parse HEAD)
+
+    # check if there are changes
+    if [ -z "$(git status --porcelain)" ]; then 
+        if [ "$OLD_HEAD" != "$NEW_HEAD" ]; then
+            echo "HEAD has changed (committed or rebased by agent). Pushing changes..."
+            git push --force origin HEAD
```
sdowell, line 331: Will this always push to the correct remote branch?

sdowell (approved):

factory/pkg/tasks/address_feedback.sh
barney-s: yes

barney-s:
this is what happens on the sandbox ... uncommitted changes


image

factory/pkg/tasks/address_feedback.sh
```diff
@@ -317,6 +318,34 @@ function runGemini {
...
+        else
+            echo "No changes to commit."
+        fi
+    else
+        echo "Changes detected in working directory, committing..."
+        git add .
+        git commit -m "Address review feedback: Apply changes"
```
sdowell, line 338: Should we do a fixup commit instead?

PR #1269, opened 2026-08-10
fix(factory): prioritize fatal quota error detection and enforce PR creation in fix tasks

barney-s:
### Summary of Changes

1. **Prioritize Fatal Quota Error Detection (`client.go` & `geminitokens.go`)**:
   - In `RunTaskResilient`, checked `IsFatalQuotaError(newData)` before `IsTransientRateLimit(newData)`.
   - Updated `IsTransientRateLimit` to explicitly return `false` if `IsFatalQuotaError(data)` is true, preventing logs that contain both `"Retrying with backoff"` and fatal quota messages (e.g. `"exceeded your current quota"`) from being falsely categorized as transient retries.
   - Added unit test cases verifying that retry attempts containing fatal quota errors are correctly identified as fatal and not transient.

2. **Enforce PR Creation in Fix Tasks (`fix.go` & `fix_issue.sh`)**:
   - In `recordPRLink` (`fix_issue.sh`), exit with code 1 (`exit 1`) if no PR URL is created/found and `NO_PR != true`.
   - In `runFix` (`fix.go`), return an error `task finished without creating a pull request` if `prNum == 0 && !noPR`, ensuring that unfinished/aborted runs are correctly marked as `Failed` in the queue rather than silently marked `Completed`.

## Triggering Issue
https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/12320#issuecomment-5245633647

the fix-issue task succeeded without creating a PR
image

### DIFF
--- a/factory/pkg/tasks/fix_issue.sh
+++ b/factory/pkg/tasks/fix_issue.sh
@@ -418,14 +418,14 @@ function recordPRLink {
     if [ -n "$pr_url" ] && [ "$pr_url" != "null" ]; then
         echo "Successfully found PR: ${pr_url}"
         echo "${pr_url}" > "$output_file"
+        popd > /dev/null
     else
         echo "Could not find PR link automatically."
-        # Don't overwrite if it already exists (unlikely here but safe)
-        if [ ! -s "$output_file" ]; then
-            echo "Could not find PR link automatically." > "$output_file"
-        fi
+        echo "Could not find PR link automatically." > "$output_file"
+        popd > /dev/null
+        echo "Task finished without creating a pull request." >&2
+        exit 1
     fi
-    popd > /dev/null
 }
 
 # Main execution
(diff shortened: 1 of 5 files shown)


factory/pkg/tasks/fix_issue.sh
```diff
@@ -418,14 +418,14 @@ function recordPRLink {
     if [ -n "$pr_url" ] && [ "$pr_url" != "null" ]; then
         echo "Successfully found PR: ${pr_url}"
         echo "${pr_url}" > "$output_file"
+        popd > /dev/null
     else
         echo "Could not find PR link automatically."
```
sdowell: nit: can use `echo "..." | tee "${output_file}"` instead of repeated echo

sdowell (approved):

factory/pkg/tasks/fix_issue.sh
barney-s: Updated to use `tee` in 4e0331c8. Thanks!

PR #1272, opened 2026-08-12
refactor: use custom flag type for parsing repo

sdowell:
This replaces the need for parsing the flag in the run function with a custom flag type. This is in preparation for a greater refactor to refactor the function signature of runWatch.


barney-s (approved):
/lgtm

thanks @sdowell

PR #1274, opened 2026-08-12
refactor: pass in flag structs to runWatch

sdowell:
This shrinks down the list of arguments for the runWatch function and other helper functions, and removes direct access to the rootFlags variable in watch.go. This prepares the files to be relocated to another package.


barney-s (approved):
/lgtm
/approve

PR #1276, opened 2026-08-13
feat: add observedGeneration to overseer CRD

sdowell:
Also fixes a couple minor bugs:
- OverseerStatus was not set to Active on initial sandbox create
- apimachinery expects unstructured objects to use list primitives as interface{} rather than string. the string slice causes a panic if the object is deep copied


barney-s (approved):
/lgtm
/approve

PR #1278, opened 2026-08-13
fix: overseer fails to restart Pods after sandbox update

sdowell:
This fixes a bug in overseer where it did not have permission to watch Pod objects, so the caching delegating controller-runtime client error looped when trying to hydrate its Pod cache. This change does two key things:

- Update the overseer RBAC to follow POLP for Pod objects. It only needs to Delete - not Get/List. Also removes the unused Event RBAC.
- Updates the controller logic to do a simple Delete without a preceding Get. This optimizes the number of API calls and removes the need for the Get/List Pod permissions.


sdowell:
chatted offline - it's actually currently intended behavior to not restart the Pod on every overseer spec update. I'll follow up with a different PR to pull out the Pod recreation code to make the intent more clear

Long term we want to do a follow up that drains the child sandbox on an overseer spec update (similar to the behavior for the [upgrade script](https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/overseer/scripts/update-overseer-factory.sh))

PR #1279, opened 2026-08-13
refactor: move shared helper functions to common pkg

sdowell:
This moves helper functions that are shared by multiple factory subcommands into a common package. This will make it easier to break out the subcommands into their own dedicated packages.


barney-s (approved):
thank you @sdowell 

/lgtm
/approve

PR #1280, opened 2026-08-13
refactor: create shared constants for factory

sdowell:
This consolidates shared constants to a common package so that they can
be shared instead of repeated/duplicated. This just starts with a few
constants but can be used as a starting point for other constants.


barney-s (approved):
/lgtm
/approve

barney-s:
perfect thanks.

PR #1298, opened 2026-08-17
fix: pr inactivity sent repeated messages

sdowell:
The pr inactivity staleness check did not properly check if the PR already had the staleness label/comment before posting a new comment.


barney-s (approved):
Doesn’t also Apple the overseer/stop label ?

PR #1308, opened 2026-08-19
refactor: move checkRepo to helper method

sdowell:


factory/pkg/commands/watch/watch.go
```diff
@@ -61,313 +107,275 @@ func (w *Watcher) Run(ctx context.Context) error {
...
+		if w.IssueMode != "disabled" {
+			w.queueIssueTasks(ctx, w.ghClient, w.kubeClient, w.cfg, issues, w.processedIssues, refIssues, w.targetAssignee, w.allBotUsers, w.incomingDir, w.processingDir, w.processedDir, w.triggerLabel)
 		}
-	}
 
-	checkRepo()
+		// Process PRs assigned to the bot in the fast cycle
+		if len(fastPRIssues) > 0 {
+			klog.Infof("Processing %d assigned PRs in fast cycle...", len(fastPRIssues))
+			w.processPRs(ctx, w.ghClient, w.kubeClient, w.cfg, fastPRIssues, w.processedPRs, w.allBotUsers, w.githubLogin, w.incomingDir, w.processingDir, w.processedDir, w.triggerLabel)
```
barney-s, line 366: in general we should use w. vars inside the method instead of passing. lets do it in a next pr

barney-s (approved):
/lgtm
/approve

barney-s (approved):

factory/pkg/commands/watch/watch.go
sdowell: +1 that's what I had planned for the next PR, just wanted to break it up so the diff wasn't too big

PR #1309, opened 2026-08-19
refactor: use watcher params instead of function args for helpers

sdowell:


factory/pkg/commands/watch/github_helpers.go
```diff
@@ -433,25 +432,27 @@ func (w *Watcher) selectUserForTask(ctx context.Context, ghClient *githubv39.Cli
...
-			if err == nil {
-				for _, a := range issue.Assignees {
-					assignee := a.GetLogin()
-					if assignee != "" {
-						inPool := false
-						for _, u := range roleCfg.Users {
-							if strings.EqualFold(u, assignee) {
-								inPool = true
-								break
+			if w.ghClient != nil {
```
barney-s: seems defensive ? was it needed earlier and we did not have it ?

factory/pkg/commands/watch/scan_pr.go
```diff
@@ -133,43 +131,46 @@ func (w *Watcher) processPRs(ctx context.Context, ghClient *githubv39.Client, ku
...
-				if rc, err := github.ListAllReviewComments(ctx, ghClient, w.Repo.Owner, w.Repo.Repo, num, r.GetID()); err == nil {
+				if rc, err := github.ListAllReviewComments(ctx, w.ghClient, w.Repo.Owner, w.Repo.Repo, num, r.GetID()); err == nil {
 					revCommentsMap[r.GetID()] = rc
 				}
 			}
 		}
 
-		state := processedPRs[num]
+		var state prWatchState
+		if w.processedPRs != nil {
```
barney-s: interesting check

barney-s (approved):
oh man .... this caused a headache :) 

/lgtm
/approve

factory/pkg/commands/watch/github_helpers.go
sdowell: agreed it seems a bit overly defensive, I don't think it seems necessary but gemini decided to do this with the refactor

sdowell: i went ahead and reverted the nil checks

PR #1312, opened 2026-08-21
refactor: extract task bash scripts into envd package with unit tests

sdowell:


barney-s (approved):
thanks @sdowell 
/lgtm
/approve

PR #1313, opened 2026-08-21
update greenfield instructions to label ready for human

barney-s:


barney-s:
this has been fixed by a deterministic approach #1314

PR #1317, opened 2026-08-25
fix: add guardrails to shared agent prompt

sdowell:
This adds several guardrails to the shared agent prompt. The guardrails are added here so they apply to all built-in and custom agent chores.

These are primarily intended to prevent the agent from performing actions that conflict with other parts of the system, for example resulting in assignment/label flapping.

### DIFF
--- a/factory/pkg/tasks/run_agent.txt
+++ b/factory/pkg/tasks/run_agent.txt
@@ -1,5 +1,7 @@
 {{ .AgentPrompt }}
 
 ---
-IMPORTANT:
-Do not commit your changes, push to the remote repository, or create a Pull Request. The surrounding system will automatically commit, push, and create/update the Pull Request for you. Just make the necessary code changes and exit.
+IMPORTANT SAFETY GUARDRAILS & SYSTEM RULES:
+1. Do not commit your changes, push to the remote repository, or create a Pull Request. The surrounding system will automatically commit, push, and create/update the Pull Request for you. Just make the necessary changes and exit.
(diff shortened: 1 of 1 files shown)


factory/pkg/tasks/run_agent.txt
```diff
@@ -1,5 +1,9 @@
 {{ .AgentPrompt }}
 
 ---
-IMPORTANT:
-Do not commit your changes, push to the remote repository, or create a Pull Request. The surrounding system will automatically commit, push, and create/update the Pull Request for you. Just make the necessary code changes and exit.
+IMPORTANT SAFETY GUARDRAILS & SYSTEM RULES:
+1. Do not commit your changes, push to the remote repository, or create a Pull Request. The surrounding system will automatically commit, push, and create/update the Pull Request for you. Just make the necessary changes and exit.
+2. Respect Stop Labels: You must NEVER remove, delete, or modify `overseer/stop` (or any `*/stop`) labels on any Pull Request or Issue. A stop label indicates automated processing was paused due to inactivity, review limits, or human maintainer intervention. Treat stopped items as paused and leave them untouched.
+3. Approval & Merging Labels: You must NEVER approve, LGTM, or add approval/merging labels (such as `lgtm`, `approved`, or `automerge`) to Pull Requests. These actions require human maintainer review and approval.
+4. Closed Pull Requests: Do not work on closed Pull Requests.
```
barney-s: Ah now i know why we need 1.
We use outer shell script to create the PR.

lets keep 1, 2. 
And remove 3,4
for 5 we do not need to call out watch daemon. Just smiple NEVER add, remove ... Issues.

sdowell: Done

barney-s:
lgtm

PR #1318, opened 2026-08-26
fix(watch): simplify bot unassignment and prevent flapping on watcher restart

sdowell:
Why lastSHA tracking was fragile on startup:
Previously, loadProcessedTasks populated state.lastSHA by parsing CommitSHA from historical task YAML files (task-pr-<num>-*.yaml) in the processed queue directory. Because os.ReadDir returns files in alphabetical order, state.lastSHA was overwritten by whichever task file sorted last alphabetically (e.g., a pr-review task from commit A would overwrite a pr-comments task from commit B), populating state.lastSHA with an arbitrary historical commit rather than the live PR head commit. Furthermore, state.unassignedSHA was purely in-memory and reset to "" on startup. On the very first scan cycle after a restart, shouldUnassignStaleBot observed a mismatch between the stale state.lastSHA loaded from disk and the live headSHA, falsely concluding that a new commit had been pushed while a bot was assigned. This caused the watcher to immediately and spuriously unassign the bot on restart, leading to assignment flapping.

How this change resolves the issue:
Instead of trying to infer commit transitions at the top of each loop using fragile disk-loaded commit deltas, this change shifts bot unassignment to a clear, state-driven model and removes state.lastSHA entirely:

1. State-Driven Bot Unassignment:
   - Moves bot unassignment to the isReadyForHuman reconciliation block in scan_pr.go. When a PR is determined to be ready for human review (clean CI, review comments addressed, automated review satisfied, and no merge conflicts), any assigned bot user is automatically unassigned.
   - Eliminates fragile commit-delta unassignment logic (shouldUnassignStaleBot) that previously unassigned bots at the start of a scan cycle before assessing if tasks were needed.

2. Dedicated CI Investigation Tracking:
   - Introduces lastInvestigatedSHA to prWatchState and populates it from task-pr-<num>-investigate.yaml in parseProcessedPRTask.
   - Uses lastInvestigatedSHA specifically to gate CI failure investigation retries without overloading generic SHA state.

3. Clean up Obsolete State and Legacy Code:
   - Removes state.lastSHA and unassignedSHA from prWatchState. Each task type (review, comments, iterate, investigate) now tracks commit gating through its own dedicated SHA field (lastReviewedSHA, lastCommentAddressedSHA, lastIteratedSHA, lastInvestigatedSHA).
   - Removes obsolete overseer/giving-up label removal logic, which was replaced by the standardized overseer/stop mechanism.
   - Removes shouldUnassignStaleBot and its unit tests.

4. Test Coverage:
   - Updates queue_test.go to verify lastInvestigatedSHA and ensure task SHAs load properly from disk without populating lastSHA.
   - Adds TestProcessPRs_UnassignOnReadyForHuman to scan_pr_test.go to test that bot assignees are removed when PRs become ready for human review.


barney-s (approved):
/lgtm

we can check and see if not-giving up is ok

PR #1321, opened 2026-08-27
chore: sync kcc workflows with upstream

sdowell:
This updates the kcc workflows to the latest version of their upstream definition (which should be treated as the source of truth).


barney-s (approved):
/lgtm

PR #1326, opened 2026-09-01
docs: proposal for new watch architecture

sdowell:
This change adds a design doc which proposes a new architecture for the factory watch command. The primary goal is to reduce system latency by breaking up the synchronous watch loop into several asynchronous subcontroller goroutines.


barney-s:
please update the doc based on our discussions

PR #1339, opened 2026-09-05
feat(watch): add TaskQueueManager with write-through disk persistence

sdowell:
Introduce TaskQueueManager in factory/pkg/commands/watch/concurrency to replace ad-hoc filesystem scanning and repetitive disk parsing with a thread-safe, in-memory queue management layer backed by write-through disk persistence and structured journal logging.

Key Changes:

1. Thread-Safe In-Memory Queue (concurrency.TaskQueueManager):
   - Implements thread-safe in-memory maps for incoming, processing, and processed tasks protected by sync.RWMutex.
   - Rehydrates state from disk on startup via LoadFromDisk() with modTime fallback.
   - Persists tasks to incoming/, processing/, and processed/ directories atomically via writeTaskAtomically (temp file write + atomic rename) to eliminate file corruption and partial reads.
   - Records task lifecycle transitions (Created, Completed, Failed) to an append-only journal.jsonl log.

2. Priority-Based Scheduling & Fair-Share Claiming:
   - Implements ClaimNextTask with multi-factor sorting across priority rankings (critical > urgent > important > high > medium > low), phase order (1 to 4), and FIFO enqueue timestamps.
   - Atomically transitions claimed tasks from incoming to processing and records StartedAt execution timestamps.

3. Complete Task Lifecycle API:
   - Adds Enqueue with automatic deduplication, CompleteTask, FailTask, and task cancellation helpers (RemoveTask, RemovePendingTasksForNumber).
   - Automatically migrates task logs from logs/processing/ to logs/processed/ upon completion or failure.
   - Adds HasActivePRTask to detect in-flight tasks for a given pull request without disk traversal.

4. Watcher & HTTP Server Integration:
   - Integrates queueMgr into Watcher and refactors task runners (runner.go), scanners (scan_pr.go, scan_issue.go), and chore schedulers (chores.go) to coordinate state through queueMgr.
   - Optimizes GET /api/v1/queue in server.go to serve responses directly from queueMgr.GetQueueResponse(), eliminating disk I/O bottlenecks on API requests.

5. Test Coverage & Concurrency Verification:
   - Adds concurrency/queue_manager_test.go covering atomic writes, crash recovery, priority sorting, lifecycle state transitions, deduplication.
   - Adds TestTaskQueueManager_ConcurrentStress to verify thread safety under high-concurrency contention across simulated workers.
   - Updates watch tests to use queueMgr.


factory/pkg/commands/watch/concurrency/priority_queue.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Len returns the total number of tasks in the priority queue.
+func (pq *TaskPriorityQueue) Len() int {
+	return len(pq.byFilename)
+}
+
+// Get retrieves a task by its filename, returning whether it was found.
+func (pq *TaskPriorityQueue) Get(filename string) (*api.QueueTask, bool) {
+	if node, ok := pq.byFilename[filename]; ok {
```
barney-s, line 77: Doesnt this DS require lock ?

factory/pkg/commands/watch/concurrency/priority_queue.go
```diff
@@ -0,0 +1,215 @@
...
+		heap.Fix(&pq.ready, node.index)
+	}
+	return true
+}
+
+// HasActivePRTask reports whether any task in the queue belongs to the specified PR number,
+// either by task metadata or by filename prefix.
+func (pq *TaskPriorityQueue) HasActivePRTask(prNumber int) bool {
+	prefix := fmt.Sprintf("task-pr-%d-", prNumber)
+	for fn, node := range pq.byFilename {
```
barney-s, line 174: same .. dont we need to protect byFilename ?

factory/pkg/commands/watch/concurrency/queue_manager.go
```diff
@@ -0,0 +1,775 @@
...
+		}
+		return nil
+	}
+
+	return fmt.Errorf("task %s not found in incoming queue: %w", filename, os.ErrNotExist)
+}
+
+// GetQueueResponse constructs the full QueueResponse payload served by the HTTP API directly
+// from thread-safe in-memory maps in sub-millisecond time.
+func (m *TaskQueueManager) GetQueueResponse() api.QueueResponse {
```
barney-s: may be move this (and http related) methods to a separate file. Not sure whats golang convention on splitting type methods across files.

factory/pkg/commands/watch/runner.go
```diff
@@ -206,84 +192,28 @@ func (w *Watcher) runSingleTask(ctx context.Context, taskFilename string, t *api
...
-			filesInProcessing++
-		}
-	}
+	}()
 
 	actionsTaken := 0
 	activeSandboxesInCycle := make(map[string]bool)
 
-	for _, item := range tasksToRun {
+	for {
```
barney-s, line 216: are we changing behavior to for ever try ?

factory/pkg/commands/watch/runner.go
```diff
@@ -297,102 +227,96 @@ func (w *Watcher) runTasks(ctx context.Context) {
...
-		task := item.Task
+		// Pop next fair-share candidate
+		filename, task, err := w.queueMgr.ClaimNextCandidate()
+		if err != nil {
+			klog.Errorf("Failed to claim next candidate task: %v", err)
+			break
+		}
+		if task == nil {
+			// No more eligible tasks ready in incoming queue
+			break
```
barney-s, line 246: i think this would cause the loop to break ?

factory/pkg/commands/watch/concurrency/priority_queue.go
sdowell: In the current design it is guarded by the lock on the TaskQueueManager (task queue manager is the only thing that directly accesses the priority queue). We could add another layer of locks to the priority queue - wdyt?

factory/pkg/commands/watch/runner.go
sdowell: Not in this iteration - here it pops each item off the queue and if it's not ready to be worked on it is added back to the queue in a deferred function. Therefore a given item won't be visited more than once per loop, and the loop will exit once the queue is empty.

I opted for the deferred re-enqueue specifically so this won't loop indefinitely, since this is still a blocking part of the synchronous watch loop. Once this is split into its own goroutine it can potentially be a forever try where it revisits the same items (e.g. after a cooldown period)

factory/pkg/commands/watch/runner.go
sdowell: Yes, this exits the loop once the queue is empty

factory/pkg/commands/watch/concurrency/queue_manager.go
sdowell: I think that adheres to convention and sounds good, I'll move this to another file

factory/pkg/commands/watch/concurrency/priority_queue.go
barney-s: later

barney-s:
lets try it out.
thanks @sdowell

PR #1355, opened 2026-09-10
feat: run task dispatcher as its own goroutine

sdowell:
This moves the task dispatcher to its own separate goroutine so that it is continuously running instead of being blocked on other operations, such as scanning the github repo issues/PRs or performing sandbox GC.

This is intended to improve the latency of picking up new tasks by decoupling the task dispatch from other long running operations. A new sandbox lease primitive is added to grant ownership of a sandbox to a given task runner.

New functionality is also added to adopt already running tasks on startup, and acquire a lease for those already running tasks until the task completes.


factory/pkg/commands/watch/queue.go
```diff
@@ -187,7 +188,20 @@ func (w *Watcher) recoverStuckTasks(ctx context.Context) {
 					if w.kubeClient != nil && sandboxName != "" {
 						running, err := isSandboxTaskRunning(ctx, w.kubeClient, w.Namespace, sandboxName)
 						if err == nil && running {
-							klog.Infof("Task %s is still actively running in sandbox %s. Leaving in processing.", f.Name(), sandboxName)
+							if w.sandboxLocks.TryAcquire(sandboxName, f.Name()) {
```
barney-s: what about the case where we restart the watch (upgrade or periodic) and the lock registry is fresh ?
we were running a task and we restarted.

factory/pkg/commands/watch/queue.go
```diff
@@ -219,6 +233,86 @@ func (w *Watcher) recoverStuckTasks(ctx context.Context) {
...
+		select {
+		case <-monitorCtx.Done():
+			if monitorCtx.Err() == context.DeadlineExceeded {
+				klog.Warningf("Adopted task %s in sandbox %s timed out after %s", taskFilename, sandboxName, w.TaskTimeout)
+				if sandboxName != "" && w.kubeClient != nil {
+					manager := k8s.NewManager(w.kubeClient)
+					_ = manager.DeleteSandbox(ctx, w.Namespace, sandboxName)
+				}
+				_ = w.queueMgr.FailTask(taskFilename, t, "adopted task timed out")
+				if t.Type == api.TypePRComments && w.cfg != nil {
```
barney-s, line 277: is this logic also done else where ? 
if so some duplication ?

barney-s: runSingleTask() .. no need to force it .. but if there is significant overlap we should consider refactoring

factory/pkg/commands/watch/runner.go
```diff
@@ -199,6 +208,53 @@ func (w *Watcher) runSingleTask(ctx context.Context, taskFilename string, t *api
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			w.runTasks(ctx)
+			// Reset the ticker after running tasks to ensure we wait the expected interval
+			ticker.Reset(cfg.interval)
```
barney-s, line 252: just want to make sure we are honoring maxsandbox limits.
since we are more aggresively draining tasks. making sure we dont overwhelm :)

factory/pkg/commands/watch/runner.go
```diff
@@ -9,11 +9,18 @@ import (
...
 
+	"k8s.io/klog/v2"
+
 	"github.com/gke-labs/gemini-for-kubernetes-development/factory/pkg/commands/watch/api"
 	"github.com/gke-labs/gemini-for-kubernetes-development/factory/pkg/k8s"
-	"k8s.io/klog/v2"
 )
 
+const (
+	DefaultDispatcherInterval = 30 * time.Second
```
barney-s, line 19: can be a bit higher

factory/pkg/commands/watch/queue.go
sdowell: Yeah this mirrors the logic in runSingleTask for the goroutine that starts running the new task and waits for it to complete (until w.TaskTimeout).

After thinking about it though I don't think we ever actually hit that code path before - the "task timeout" is configured to 3 hours and the "watch timeout" is configured to ~30 minutes. So the watch timeout is hit and the process exits before ever hitting the task timeout.

Therefore this may be more of a bugfix, which I fear could have unintended side effects. WDYT? Would it be better to break this out into a separate change?

factory/pkg/commands/watch/runner.go
sdowell: Did you have a certain duration in mind? I was actually thinking this could be lower (0-10s) but went with the 30s to mirror the [current minimum interval of 30s between running new tasks](https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/factory/pkg/commands/watch/watch.go#L246)

A lower interval would increase responsiveness, running new tasks sooner when they hit the task queue. WDYT?

factory/pkg/commands/watch/runner.go
sdowell: Yeah we still honor the max sandbox limits, but agree we will be having higher utilization now that tasks are dispatching more quickly - we might need to tune down the max sandbox limit if we run into capacity issues

factory/pkg/commands/watch/queue.go
barney-s: removing existing code .. yes ...

factory/pkg/commands/watch/runner.go
barney-s: i think we need a better metric ... lets add something along the line of taskrate (separate PR)

factory/pkg/commands/watch/runner.go
barney-s: just general concern about capacity 1!!

barney-s:
lets iterater. this looks good.

PR #1360, opened 2026-09-11
refactor(watch): extract task dispatcher into its own package

sdowell:
The task dispatch loop, task execution, and stuck-task recovery were all implemented as methods on Watcher, which left the dispatch logic entangled with GitHub scanning, Kubernetes clients, and CLI flag parsing. This made the dispatcher impossible to unit test in isolation and gave Watcher an unclear set of responsibilities.

Move dispatching into a new pkg/commands/watch/dispatcher package:

- Dispatcher owns the dispatch loop, concurrency limits, sandbox leases, and queue state transitions.
- CLIRunner implements the new TaskRunner interface. Process spawning and executable resolution are injectable struct fields, replacing the package-level execCommand var.
- SandboxService and TaskCoordinator interfaces keep the dispatcher free of any direct k8s or GitHub dependency. The watch package supplies adapters in dispatcher_deps.go.
- Recover() takes over stuck-task recovery and adopted-task monitoring from Watcher.recoverStuckTasks and Watcher.monitorAdoptedTask. Supporting disk-aware helpers (SyncProcessingFromDisk, ProcessingTasks, RequeueUntrackedProcessingFiles) live in concurrency, since they need the queue manager's unexported directory fields.
- Drain detection moves from watch.isDoNotProcess to concurrency.IsDrainMode, so the dispatcher no longer needs a queue dir.

With adopted-task monitors now tracked by the dispatcher's WaitGroup, the Watcher no longer needs one of its own; Watcher.Wait delegates to the dispatcher.

Two intentional behavior changes:

- Task execution skips context.WithTimeout when TaskTimeout is 0. It previously created an already-expired context, killing the task immediately.
- Timeout cleanup deletes the sandbox name resolved at claim time rather than re-deriving it, so a task can no longer delete the wrong sandbox if resolution is non-deterministic.

Also updates design/watch-subcontrollers-architecture.md to match: marks Phase 2 complete, drops the 500ms ticker and wakeup channel from the dispatch design, and replaces the dual-WaitGroup model with the subcontroller-owned WaitGroups that were actually implemented.


barney-s:
i went through all the files (except some test files). ... seems ok ..
But we need to test in autopush ...

especially upgrades.

PR #1361, opened 2026-09-11
fix(tasks): require a GitHub closing keyword when linking issues

sdowell:
The PR hygiene instructions offered "Issue #NNNN" and "Fixes #NNNN" as equivalent, but only the latter is a closing keyword. Agents often wrote prose like "resolves Issue #NNNN", which GitHub does not parse as a link, so the PR merged without closing its issue.

The issue scanner's duplicate guards are all keyed on the linked PR being open, so they release the moment it merges and the issue is dispatched again. That produced #1359, a duplicate of the merged #1357.

Require the keyword to be followed immediately by the issue reference, and call out the near-miss phrasings as non-linking.


factory/pkg/tasks/address_feedback.txt
```diff
@@ -12,7 +12,9 @@ Use good PR hygene:
 * **Avoid committing or pushing large binary files.** Build artifacts (like the `manager` binary in many controller projects) should be cleaned up before committing or added to `.gitignore`.
 * **Cleaning up:** If you run build commands (like `make`, `go build`, `go test`), ensure you clean up any generated binaries or artifacts before committing. Use `make clean` if available, or manually remove them.
 * **Do NOT approve, LGTM, or label PRs**: You must **never** approve, LGTM, or add approval/merging labels (such as `lgtm`, `approved`, or `automerge`) to any Pull Requests. These actions require human OWNER review and approval.
-* Reference the issue you are solving in the PR description, like "Issue #1234" or "Fixes #1234".  Put this in the commit body, not the first line.
+* **Link the issue with a GitHub closing keyword.** The PR description MUST contain a line of exactly the form `Fixes #1234` (or `Closes #1234` / `Resolves #1234`), where the keyword is followed immediately by `#` and the issue number, with no other words in between. Put it on its own line.
```
barney-s: lets stick to one. Fixes if that is standard.

sdowell: agreed, done

PR #1365, opened 2026-09-11
fix(watch): stop re-triggering issues that already have an open PR

sdowell:
The watcher commented "AI Factory started fixing this issue in a sandbox" on k8s-config-connector#9259 and spawned a duplicate sandbox even though PR #12689 ("Fixes #9259") had been open for nine days. Two independent duplicate-suppression mechanisms failed at the same time.

1. Timeline check was unpaginated. ListIssueTimeline was called with nil options, so only the first 30 events were inspected. Issue #9259 has 46 events and the cross-reference to the open PR is event 45, so the check only ever saw an older closed PR and a plain issue reference.

   listAllIssueTimeline now follows pagination at per_page=100 (capped at 20 pages) and reports whether the read completed. Callers only trust a negative answer from a complete timeline; otherwise they fall back to the Search API, which previously only ran when the timeline errored.

2. The open PR cache was empty. The process restarted into a GitHub rate limit window, both attempts to list open PRs returned 403, and the scan proceeded with an empty referencedIssues map, which is indistinguishable from "no PR references this issue".

   canQueueIssueTasks now fails closed: hasPRs is only set after a successful listing, and no issue work is queued until then.

### DIFF
--- a/factory/pkg/commands/watch/github_helpers.go
+++ b/factory/pkg/commands/watch/github_helpers.go
@@ -105,24 +105,63 @@ func getMissingLabelsForPR(prLabels []*githubv39.Label, refIssues []*githubv39.I
 	return allMissingLabels
 }
 
-func hasLinkedPR(ctx context.Context, client *githubv39.Client, owner, repo string, issueNum int) (bool, error) {
-	// 1. Try timeline check (quick and standard)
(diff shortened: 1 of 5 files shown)


barney-s (approved):
/lgtm

PR #1371, opened 2026-09-14
docs(repo-agent): factory-CLI migration design doc

barney-s:
## Summary

Design doc for refactoring `repo-agent/` to consume the **factory CLI** as its agent engine, the same way overseer does — while keeping repo-agent's differentiator: a hosted, multi-tenant, **user-OAuth-identity** service.

Key points:

- **Contract**: process invocation only (bundle the factory binary in the image, exec `factory fix / pr review / watch` with `.factory.cfg` + `FACTORY_CONFIG`); no Go-level imports — same as overseer.
- **Identity**: tenant bootstrap normalizes the user's OAuth token into the `factory-user` Secret per namespace (mirror of overseer's `reconcileFactorySecret`, sourced from OAuth instead of bot PATs).
- **Keep**: OAuth/API/UI shell, RepoWatch CRD, dev sandboxes, configdir/syncer. **Delete**: SandboxTask+taskrunner, repo-sandbox agent subcommands, pkg/tasks, pkg/llm, gemini-stream-processor.
- **Phases**: (1) fix-issue path → (2) review path + UI rewire → (3) factory-watch adoption → (4) deletions. Each phase cuts its task type over completely.
- **Hard constraint**: `factory/` and `overseer/` remain unaffected — all gaps closed repo-agent-side (e.g. triage as an `.agents/` definition via `factory agent create --local`). One carve-out: the Overseer admin dashboard + token-usage views hosted in repo-agent's API/UI must keep working throughout.

Follow-up PRs will implement Phase 1 onward, one at a time.


repo-agent/docs/design/factory-cli-migration.md
```diff
@@ -0,0 +1,284 @@
...
+  (building the binary from monorepo source, as overseer does, is consumption, not
+  change). Every gap is closed on the repo-agent side or deferred.
+- **Breaking changes inside repo-agent are allowed at every phase** — its API routes,
+  CRDs, UI, images, and manifests can change or disappear without compatibility
+  shims. The only exception: repo-agent's API/UI also hosts the *Overseer admin
+  dashboard* and token-usage views (`pkg/api/handlers_overseer.go`,
+  `handlers_usage.go`, the Overseer/TokenUsage UI views) — those serve the overseer
+  stack and must keep working throughout.
+
+What differentiates the three after the migration:
```
sdowell, line 29: Assuming the key difference is the identity that the agent runs as, this has me wondering if we could take it a step further and have repo-agent just invoke `factory watch` with a different set of args/identities/etc. WDYT?

It would be a nice side effect to share as much code as possible between the two operational models, but I suspect there may be other difference that aren't captured in this table

sdowell (approved):

PR #1372, opened 2026-09-14
feat(repo-agent): factory-CLI migration groundwork (Phase 1a)

barney-s:
## Summary

First implementation PR of the factory-CLI migration (design doc: #1371). Groundwork only — **no behavior change to task dispatch yet**.

- **Bundle the factory binary** into the repowatch-controller image, built from monorepo source exactly like overseer's image does. Process invocation remains the only repo-agent↔factory contract; no Go-level imports.
- **Reconcile a `factory-user` Secret per tenant namespace** from the repowatch controller, with the keys factory expects (`GITHUB_TOKEN`/`GITHUB_LOGIN`/`GITHUB_EMAIL`/`GEMINI_API_KEY`). The GitHub token follows the existing precedence (`manual_pat` > `oauth_pat` > `pat`) and is re-read from the secret each reconcile so OAuth refreshes propagate; login/email come from the authenticated user already fetched in `Reconcile`; the Gemini key from `gemini-vscode-tokens`. This mirrors overseer's `reconcileFactorySecret`, sourced from user OAuth instead of bot PATs.

`factory/` and `overseer/` are untouched.

## Test
- `go build ./...` and `go test ./pkg/controllers/repowatch/` pass; new unit tests cover secret create/update, token precedence, email fallback, and the missing-token error path.

Next PR: switch the issue-fix dispatch path to `factory fix --detached`.


repo-agent/pkg/controllers/repowatch/factory_secret.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := r.Get(ctx, types.NamespacedName{Name: k8s.GeminiSecretName, Namespace: repoWatch.Namespace}, geminiSecret); err == nil {
+		for _, key := range []string{"gemini", factoryKeyGeminiAPIKey} {
+			if v, ok := geminiSecret.Data[key]; ok && len(v) > 0 {
+				data[factoryKeyGeminiAPIKey] = v
+				break
+			}
+		}
+	}
+
+	desired := &corev1.Secret{
```
sdowell, line 90: nit: suggest breaking out the logic for fetching all of the secrets and constructing the desired state into a helper function

sdowell (approved):

repo-agent/pkg/controllers/repowatch/factory_secret.go
barney-s: deferring for a longer refactor at the end

PR #1373, opened 2026-09-14
feat(watch): sandbox GC as its own reconciler

sdowell:
watch/sandbox.go mixed two jobs as Watcher methods: answering per-task questions ("which sandbox does this task run in?", "is it busy?") and sweeping the cluster for sandboxes to suspend or delete. Both ran inline from checkRepo(), so every scan cycle paid for pod lists, envd probes and GitHub confirmations before it could queue any work.

Split them into pkg/commands/watch/sandbox:
- Service: the synchronous lookups and probes the scanners and dispatcher need.
- Reconciler: an autonomous goroutine doing the slow work, with a 30s ticker for task state and evicted pods and a 5m ticker for garbage collection. checkRepo() no longer touches the cluster.

The reconciler sees queue state only through SandboxLockRegistry, never the TaskQueueManager, so collection cannot mutate the queue and always skips leased sandboxes — hence SuspendIdleSandboxesExcept in pkg/sandbox.

EntityStateCache moves open PR and issue state off watchState into concurrency: scans publish what they saw, and collection only confirms unknown entities against GitHub, skipping entirely until the cache is warm.

Also: Watcher.Run joins subcontrollers with a local WaitGroup; dispatcher_deps.go becomes subcontrollers.go, where renaming the SandboxService methods to match (ResolveSandboxName -> ResolveName, DeleteSandbox -> Delete) lets *sandbox.Service implement it directly and deletes the adapter; KubernetesClient.Clientset widens to kubernetes.Interface so the reconciler's pod paths can be faked in tests.

Step 3a of the subcontroller rearchitecture — the chore scheduler and scanners still run inside checkRepo().


factory/design/watch-subcontrollers-architecture.md
```diff
@@ -207,15 +207,21 @@ To guarantee maintainability and eliminate circular dependencies, subcontrollers
...
+* **Location**: Dedicated `watch/sandbox` package. It exposes two collaborators so that the cheap per-task lookups and the expensive sweeps are not entangled:
+  * `sandbox.Service` — point lookups and probes (`ResolveName`, `IsTaskRunning`, `IsTaskCompleted`, `CountRunningTasks`, `Delete`) shared by the scanners and the dispatcher. It is the only place that knows how a task number maps onto a sandbox name.
+  * `sandbox.Reconciler` — the autonomous goroutine, built via `sandbox.New(Config, Deps)`.
+* **Cadence**: Two independent tickers, because the two workloads have very different costs:
+  * `Interval` (default **30s**) drives `ReconcileOnce`: deleting evicted pods and refreshing the recorded task state of sandboxes that still claim to be running. These are namespace-scoped cluster calls with a fixed cost.
+  * `GCInterval` (default **5m**) drives `CollectGarbage`. Collection confirms every candidate against GitHub before deleting it, so its cost scales with the number of sandboxes whose entity is missing from the cache. Running it on the fast ticker would multiply GitHub API traffic by 10x for no latency benefit, since nothing waits on garbage collection.
```
barney-s, line 215: the cost of running sandbox is low. So can be 30m or even 60m default

factory/pkg/commands/watch/sandbox/reconciler.go
```diff
@@ -0,0 +1,466 @@
...
+
+// Run reconciles sandbox state until ctx is cancelled.
+//
+// State reconciliation runs immediately and then once per Interval, while the
+// heavier garbage collection sweep runs once per GCInterval.
+func (r *Reconciler) Run(ctx context.Context) error {
+	ticker := time.NewTicker(r.cfg.Interval)
+	defer ticker.Stop()
+
+	gcTicker := time.NewTicker(r.cfg.GCInterval)
```
barney-s, line 110: would GCing every hour affect something else ?

factory/pkg/commands/watch/sandbox/reconciler.go
```diff
@@ -0,0 +1,466 @@
...
+	defer gcTicker.Stop()
+
+	r.ReconcileOnce(ctx)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			r.ReconcileOnce(ctx)
```
barney-s, line 120: is this reconcile reconciling tasks running in a sandbox ? 
meaning checking if they are completed or failed ?

factory/pkg/commands/watch/sandbox/reconciler.go
sdowell: I think the main tradeoff would be if we hit the sandbox limit, then new tasks that require a new sandbox might get starved (until the next delayed GC execution)

factory/pkg/commands/watch/sandbox/reconciler.go
sdowell: Yeah the main thing this is doing is reconciling the task annotations on the sandbox object (e.g. `sandbox.gemini.google.com/last-task-state`)

barney-s (approved):
/lgtm
/approve

barney-s:
lets try this out

PR #1376, opened 2026-09-15
fix(watch): cancel task workers under a separate ctx

sdowell:
Since the task goroutines shared a context with the parent dispatcher goroutine, cancelling the daemon context immediately killed all of the task goroutines and marked the task as failed. This change accomplishes two things:
- Handles context canceled in the worker, leaving the task in processing/ and gracefully exit
- Creates a separate worker context that can be forcibly canceled on its own after the grace period


factory/pkg/commands/watch/dispatcher/dispatcher.go
```diff
@@ -22,6 +22,9 @@ import (
 const (
 	// DefaultInterval is the interval between dispatch cycles when none is configured.
 	DefaultInterval = 30 * time.Second
+	// DefaultShutdownGracePeriod is how long Run waits for in-flight tasks to finish
+	// after its context is cancelled, when none is configured.
```
barney-s, line 26: when does this cancel happen ? 
Do we detach or delete the pod ?

factory/pkg/commands/watch/dispatcher/dispatcher.go
```diff
@@ -163,6 +194,65 @@ func (d *Dispatcher) Wait() {
...
+	}()
+
+	grace := d.shutdownGracePeriod()
+	select {
+	case <-done:
+		return
+	case <-time.After(grace):
+	}
+
+	klog.Warningf("In-flight tasks did not finish within the %s shutdown grace period. Cancelling their supervisors; the tasks stay in the processing queue and are recovered on the next run.", grace)
```
barney-s: does the task keep running in the remote sandbox ? 
If so we need to indicate that

factory/pkg/commands/watch/dispatcher/dispatcher.go
sdowell: It happens in the `drain` method below, which is [reached after the parent context is cancelled](https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/factory/pkg/commands/watch/watch.go#L66). The sandbox/pod is not stopped, just the supervisor goroutines are stopped as the `factory watch` exits.

Note: This is roughly the same behavior as before, just handling the exit more gracefully by collecting the goroutines before exiting

factory/pkg/commands/watch/dispatcher/dispatcher.go
sdowell: Yes the tasks keep running in the remote sandbox since they are invoked with `nohup`, on next startup of watcher the still running tasks will be recovered/adopted and monitored until they complete

PR #1377, opened 2026-09-15
fix: prune gemini-3.8-flash from models

sdowell:
This model is experiencing high demand and frequently fails. Removing it from the default list for now to improve reliability.


barney-s (approved):
i would put it down on the list but this is ok too ...

PR #1398, opened 2026-09-16
feat(watch): chore scheduler as its own subcontroller

sdowell:
watch/chores.go ran inside the slow PR cycle of checkRepo(), which meant a chore came due up to 5 minutes before anything noticed, every cycle re-listed and re-read .agents/ from GitHub, and any PR scan latency delayed scheduling further.

Extract it into pkg/commands/watch/chores as an autonomous Scheduler goroutine (and a standalone pass in --once mode) with two cadences split by cost:
- 30s evaluation ticker (ScheduleOnce): pure in-memory cron comparison against the cached definitions, so chores fire within 30s of coming due.
- 5m definition refresh (RefreshInterval): caches the parsed .agents/ definitions and serves the previous copy if a refresh fails, so a GitHub outage only delays picking up edits rather than stopping chores from firing.

The scheduler depends on narrow Source (ListAgentFiles, ReadAgentFile) and Queue (TaskExists, Enqueue) interfaces rather than GitHub or cluster clients. chores_state.json keeps its on-disk schema but now persists via temp-file + atomic rename so an interrupted write cannot read back as empty state and re-fire every chore.

Also introduce a repo-bound github.Client in pkg/github (holding an unexported *githubv39.Client rather than embedding it) with ListAgentFiles, ReadAgentFile, and status-based IsNotFound helpers. *github.Client satisfies chores.Source directly, eliminating the need for a watch-side adapter. Generalize Watcher.reclamationPaused to Watcher.draining so both the sandbox reconciler and chore scheduler hold off on new work during drain mode.

Step 3b of the subcontroller rearchitecture — issue and PR scanners still run inside checkRepo().


barney-s (approved):
/lgtm
/approve

PR #1414, opened 2026-09-17
feat(repo-agent): clicked reviews post as your pending review on GitHub — no in-app publish

barney-s:
## What

Reworks the review flow around `factory pr review --publish draft` (already supported by factory — no factory changes): **a clicked review posts a pending review on GitHub under the clicker's identity, and they finalize it natively there.** The in-app "Publish review" button, endpoint, and draft-copy step are removed.

### Why identity had to move with it

A GitHub pending review is visible **only to its author**. So the click-Review path now runs in the **clicker's namespace with their token** (exactly like Fix) instead of the board's discovery identity — which also makes it a genuine attributed write, consistent with the human-identity thesis: click = consent = your review, authored by you, finished on GitHub where review tooling is best.

### Mechanics

- `factorycli.ReviewOptions` gains `Publish` (default `no`, unchanged args otherwise).
- `reviewPlan{pr, executor}` replaces bare PR numbers: mailbox clicks carry the member and route to their namespace/token with `--publish draft`; trigger-label and `intake.draftReviews` reviews stay **draft-only** under the discovery identity (no GitHub writes without a click — the §5 invariant is untouched).
- With `--publish draft` factory does not print the CODE REVIEW banner, so there is nothing to harvest: completion marks the sandbox `reviewState=pending`; the board renders **"Pending on GitHub"** (needs-you) with a **Finalize on GitHub ↗** link and Re-review.
- Intake drafts (`review-ready`) remain display-only in the UI, now with a **"Review as me"** action that runs the attributed review.
- `POST /board/:b/prs/:n/publish` route + handler removed; design doc §API updated.

## Testing

Controller: `TestMailboxReviewPendingOnGithub` (completion → `reviewState=pending`, no draft, mailbox cleared) and `TestMailboxReviewLaunchAsExecutor` (launch in clicker ns, their token, `--publish draft`). API: `review-pending` stage mapping; publish handler test removed with the handler. All green; JSX balance-verified.


barney-s:
Follow-ups per review: review-pending rows now offer **Abandon review** (deletes your pending review on GitHub — only one may exist per user — clears sandbox state, pauses it; a fresh Review click relaunches via the re-review marker, and stale invocation results are never re-recorded after an abandon). The review-ready action is plain **Review** (attributed run). **Merge removed everywhere** incl. the endpoint — merging happens on GitHub; Promote is the one repo-write left on own PRs.

barney-s:
One more per review discussion (b6745618): **Fix again removed** from successful-fix rows (fix-done / pr-open / My PRs) — factory reuses the fix sandbox, so re-runs would clobber the prior candidate's workspace rather than produce true parallel candidates. fix-failed keeps a same-sandbox **Retry** (nothing shipped, clean do-over). Real candidate PRs would need per-run sandboxes on the factory side — parked.

PR #1415, opened 2026-09-17
fix(repo-agent): boards without a limits block silently blocked every launch

barney-s:
First live-test bug from #1414: clicking Review on the barney-s board did nothing — mailbox was written, controller reconciled, and exited silently at the limits gate. UI-created boards have no `limits` block, and CRD field defaults don't apply when the parent object is absent, so `MaxActive=0` vetoed every fix/review/triage launch with zero logging. Zero/absent limits now fall back to the CRD defaults (5 active / 2 per user) in code, and deferred launches log the reason. Regression test: a board with an empty LimitsSpec launches a mailbox review.


barney-s:
This PR now carries all three first-live-test fixes, found by walking one Review click through the stack:

1. **Limits defaulting** — UI-created boards (no `limits` block) silently blocked every launch (`MaxActive=0`); absent/zero now means the CRD defaults, and deferred launches log why.
2. **RBAC** — the 8b regen dropped services/pods rules with the repowatch markers; factory (running under the controller SA) could not create the sandbox's -lb Service (also the cause of the /sandbox proxy Bad Gateway). Markers added, rbac.generated.yaml regenerated, already `kubectl apply`'d to the live cluster.
3. **kubectl in the image** — factory's envd layer shells out to kubectl (sandbox lookup, port-forward); the distroless controller image had none, so sandboxes came up with daemon+envd idle and no task scheduled. Now bundled like overseer's image.

After merge+deploy, clicking Review on #1163 again (or the pending mailbox) should drive the existing sandbox end to end.

PR #1419, opened 2026-09-17
fix(repo-agent): offer Review despite a leftover paused sandbox

barney-s:
Live-test find on the Abandon flow: abandoning #1163's pending review worked end to end (review deleted on GitHub, reviewState cleared, abandoned-at stamped, sandbox paused, row honestly back to 'Review requested') — but the row had **no Review button**, because the action was gated on `!item.sandbox` and the paused sandbox still exists. Dead end: Abandon → can't review again.

The stages `open` / `review-queued` / `review-requested` only occur when no run is active or pending (`reviewing`/`review-pending`/`review-submitted` cover those), so the stage itself is the guard — the sandbox check is dropped. A fresh Review click stamps the re-review marker and the controller relaunches into the existing sandbox, exactly the abandon→review-again loop the design intends.

Issue-side Fix gating keeps its sandbox check deliberately: an issue at 'open' with a sandbox means a launch is provisioning.


barney-s:
Grew per live findings: (1) **kickoff no longer writes the trigger label** — the best-effort label turned a one-time click into a standing trigger (PR #1367 auto-relaunched after cleanup with an empty mailbox; the label my click had added was the artifact). Assignment/self-review-request remain as the claim; mailbox stays authoritative. (2) **Labeled-PR stages are honest per board type**: personal boards keep 'Review queued' (the controller really will run it); shared boards show 'needs-reviewer' (needs-you + Review action) since a label alone launches nothing post-#1418. (3) Review action no longer hidden by a leftover paused sandbox (the original fix).

PR #1420, opened 2026-09-17
feat(repo-agent): settle reviews once submitted on GitHub

barney-s:
Found immediately after the first full review cycle completed live (PR #1367: click → attributed run → pending review → finalized natively on GitHub as COMMENTED): nothing ever settles `reviewState pending → submitted`, so the row keeps shouting 'Pending on GitHub / needs you' after you've already submitted.

The controller now checks pending-marked sandboxes each reconcile **under the executor's own token** (a pending review is only visible to its author): when the executor's pending review is gone and a submitted review (APPROVED / CHANGES_REQUESTED / COMMENTED) exists, `reviewState` flips to `submitted` and the row drops to the quiet `review-submitted` stage. Cost: one ListReviews call per pending sandbox per reconcile — bounded by in-flight reviews.

Regression test with a mocked reviews endpoint. (Manual deletion of the pending review on GitHub — outside Abandon — still leaves a pending row; Abandon remains the in-app cleanup.)


barney-s:
Extended per review: **submitted is not a dead end**. (1) A re-requested review on a submitted row (GitHub clears you from requested_reviewers on submit; re-request re-adds) returns the row to review-requested/needs-you — the native follow-up signal after new pushes. (2) Submitted rows always offer **Review again** (slot is free; fresh run → new pending review). (3) An active run outranks stale reviewState. Known limit: the auto-review tier only picks up first requests — re-requests need a click for now.

barney-s:
One more live find: after clicking Review, the refetched row still showed 'Review requested / Needs you / [Review]' until factory stamped Running (~1 min) — stale feedback inviting a double click. Fixed server-side so any refresh tells the truth: **mailbox entries render their item as Starting…/working** (the click is the state), and **a sandbox with no task state renders Starting… too** (provisioning window). No kickoff buttons in either state; duplicate clicks were already harmless (single-flight) but now they're not invited.

PR #1422, opened 2026-09-17
feat(watch): issue and pull request scanners as their own subcontrollers

sdowell:
Issue and pull request scanning were two branches of checkRepo() sharing one goroutine: a 30s pass over assigned and operator-filed issues, a labelled issue sweep riding inside the slow 5m PR cycle, and the PR evaluation itself. A repository with a few dozen active pull requests pushed issue pickup out by the length of whatever GitHub was slow to answer, because all of it was serialized behind one loop.

Extract both into their own packages and goroutines, completing Phase 3 of the subcontroller rearchitecture.

pkg/commands/watch/issues, two cadences split by cost:
- 30s ownership scan over issues assigned to a bot in the pool or filed by the operator, one page each sorted by update time. This is what sets pickup latency, and it is no longer behind anything.
- 5m sweep: the paginated trigger-labelled listing, which also publishes the open-issue set to EntityStateCache for the sandbox reconciler. Running it on the fast ticker would multiply its GitHub traffic tenfold to surface issues the fast queries already carry.

pkg/commands/watch/prs, same split for the same reason. The design doc specified a single 1m ticker; evaluating one pull request costs the better part of a dozen GitHub requests, so a 1m cadence over the full set would exhaust a busy repository's hourly rate limit:
- 1m fast pass over the pull requests assigned to the bot pool, which are the ones with work in flight and therefore the only ones whose CI results, review feedback and merge state change between cycles.
- 5m sweep over everything the watcher is responsible for, which also refreshes the open PR cache. The two are alternatives within a cycle rather than additions, since the sweep's candidate set is a superset of the fast pass's. Within a cycle, candidates are evaluated on a bounded four-worker pool and deduplicated by number, so no two workers ever touch one pull request's state.

Both scanners depend on narrow Queue, Entities, Sandboxes and (for issues) UserSelector interfaces rather than on the watcher. Role-based bot selection stays behind UserSelector because it reads the factory config and can pin a task to an existing sandbox's account.

Per-entity processed state stays with its owner rather than moving to EntityStateCache as §5C proposed. The deciding question turned out to be who reads it, not how many goroutines write it: nothing outside the PR scanner reads which commit a pull request was last reviewed at, and nothing outside the issue scanner reads when an issue was last worked on. The issue scanner keeps a plain map, being single-goroutine; the PR scanner a mutex-guarded stateStore, which is all the worker pool needs. The shared cache stays reserved for state that genuinely crosses subcontrollers.

Pull requests returned by the issue queries are dropped rather than handed over, keeping the dependency graph acyclic with no channel between subcontrollers. The cold-cache guard canQueueIssueTasks provided is preserved and tightened: the issue scan fails closed when no PR listing has ever succeeded, and primes that half of the cache itself in modes where no PR scanner will.

Also adds pkg/commands/watch/conventions, a leaf package holding the vocabulary the scanners and the dispatcher's coordinator share: the label grammar (HasStopLabel, StopLabel, HasTriggerLabel, Priority), bot classification (AssignedBotUser, IsReviewerBot, IsBotReply, ShouldIgnoreUser) and the comment/reaction policy.

checkRepo() is gone, and with it watchState, queue.go, scan_issue.go and scan_pr.go; Watcher.Run is left supervising subcontroller goroutines rather than driving a cycle of its own. github_helpers.go is down to selectUserForTask and renamed user_selection.go. Mode gating moved to Watcher.issuesEnabled and Watcher.prsEnabled, beside choresEnabled.

The subcontrollers talk to GitHub through the repository-bound client rather than a raw one. Each scanner used to hold a *githubv39.Client plus an Owner and Repo it re-threaded through every call; they now hold a *github.Client, built once as Watcher.repoClient and shared with the chore scheduler and the dispatcher's task coordinator, and Owner/Repo came off both Configs. Everything the watcher does to a repository is now a method on that type - issue and PR reads, label and assignee writes, comments and reactions, check runs and statuses, file contents, merge queue state, and the timeline and linked-PR lookups that moved out of issues/timeline.go into pkg/github/timeline.go with their tests.

The paginated free functions other call sites still use - github.ListAll* and common.ListAllCheckRuns/ListAllStatuses - were inverted into one-line shims over the new methods rather than deleted. Keeping them means pr.go, fix.go and agent.go did not have to change; they are the remaining adoption debt, marked deprecated so it is visible.

Two things deliberately stayed out of pkg/github. HasIssueCommentReaction and ResolveCommentReactions remain in watch/conventions, because they mix GitHub I/O with factory bot policy - they ask ShouldIgnoreUser whose reaction it is - and moving them whole would make the low-level package import the watcher's. They now call the client for the I/O and keep only the policy. IsWorkflowDefinition, GetWorkflowCooldown and FetchWorkflowContent remain in commands/common as workflow policy, but take a *github.Client; Client.ForOtherRepo covers the one case that needs another repository, a workflow definition referenced by URL, so the raw client never has to be handed back out.


factory/pkg/commands/watch/types.go
```diff
@@ -83,16 +89,15 @@ type Watcher struct {
...
-	processedPRs     map[int]prWatchState
 	queueMgr         *concurrency.TaskQueueManager
 	sandboxLocks     *concurrency.SandboxLockRegistry
 	entityCache      *concurrency.EntityStateCache
 	sandboxes        *sandbox.Service
 	dispatcher       *dispatcher.Dispatcher
 	reconciler       *sandbox.Reconciler
 	chores           *chores.Scheduler
-	state            *watchState
+	issues           *issues.Scanner
```
barney-s: issueScanner

factory/pkg/commands/watch/types.go
```diff
@@ -83,16 +89,15 @@ type Watcher struct {
...
+	prs              *prs.Scanner
```
barney-s: prscanner

factory/pkg/commands/watch/prs/scanner.go
```diff
@@ -0,0 +1,513 @@
...
+	// Evaluation cost scales with the number of open pull requests, at roughly
+	// a dozen requests each, so this interval is what keeps a busy repository
+	// inside its hourly GitHub rate limit. Shortening it is not a free latency
+	// win: the fast pass above already covers the pull requests that have work
+	// in flight.
+	DefaultSweepInterval = 5 * time.Minute
+	// DefaultWorkers is how many pull requests are evaluated concurrently.
+	// The pool bounds the burst of GitHub requests a cycle can produce while
+	// still keeping one slow pull request from delaying the others.
+	DefaultWorkers = 4
```
barney-s: lets start with 2

Issue #1341, opened 2026-09-08
Add gemini-for-kubernetes-development to awesome-ai-plugins?


zerocodefast:
Hey, gemini-for-kubernetes-development adds Gemini CLI skills that author declarative validation for Kubernetes API resources and generate structured PR review reports, useful for anyone contributing to kubernetes/kubernetes.

We maintain [awesome-ai-plugins](https://github.com/hashgraph-online/awesome-ai-plugins). Would you be up for adding gemini-for-kubernetes-development with a PR? You can [edit the README directly](https://github.com/hashgraph-online/awesome-ai-plugins/edit/main/README.md).

Suggested entry:

- [gemini-for-kubernetes-development](https://github.com/gke-labs/gemini-for-kubernetes-development) - Gemini CLI extension that automates Kubernetes development tasks: declarative validation authoring, PR review, and SIG API Machinery issue triage.

[Submission guidelines](https://github.com/hashgraph-online/awesome-ai-plugins/blob/main/CONTRIBUTING.md).

barney-s:
**Triage:** This issue is an invitation from the maintainers of awesome-ai-plugins to add gemini-for-kubernetes-development to their curated list. Because it involves submitting a PR to an external repository, this is a non-code task affecting project visibility and documentation rather than the codebase itself. The suggested next step is for a contributor to submit the suggested entry to the awesome-ai-plugins repository and close this issue.

Suggested priority: low

PR #1424, opened 2026-09-17
fix(repo-agent): resume pass for triage — harvest clicked triages after completion

barney-s:
First live [Triage] click (issue #1341): the run completed but the row stuck at 'Triaging… / Agent working' with a 'completed' chip — the suggestions were never harvested. The mailbox entry (click-mode's only plan source) is consumed when the sandbox appears, minutes before completion, and triage had no resume pass (reviews have resumeReviews). With auto-triage off, nothing ever re-drove ensureTriage.

resumeTriages mirrors resumeReviews: triage sandboxes in the board namespace without a triaged-at stamp re-enter ensureTriage, which harvests the remembered invocation output (or relaunches after a controller restart). Regression test: completed-but-unharvested sandbox + result in memory, no mailbox, no intake → draft stored, no relaunch.

After deploy, #1341's row settles to 'Triage ready' on the first reconcile if the controller hasn't restarted (result still in memory); after a restart it re-runs once.


barney-s:
Also per review: the Triage tab is now a single-verb inbox — **no Fix button on triage-group rows** (untriaged or triage-ready). Assignment or the trigger label moves an issue into the Fix tab, where Fix lives. (Live note: issue #1341's completed-but-unharvested triage was hand-harvested on the cluster, so its row settles without a re-run; this PR's resume pass makes that automatic for all future clicks.)

barney-s:
Per review: **Publish triage** lands the last unwired verb — applies the suggested labels and posts the assessment comment on the issue **under the clicker's token** (human identity, one click), then the row settles to a quiet 'Triaged' stage. The suggestions display also moves out of the Title cell: a **Suggestions** toggle beside the action buttons expands a full-width sub-row (the actions column itself is too cramped for YAML). Board action vocabulary is now: Fix · Retry · Triage · Publish triage · Review · Review again · Finalize↗ · Abandon · Promote.

PR #1428, opened 2026-09-17
style(repo-agent): one Status column — stage words, attention colors

barney-s:
Per review: Stage + Attention were two columns splitting one signal. Merged into a single **Status** chip: the stage's words (which already name both the progress — Fixing…, Triaging… — and the needed decision — Triage ready, Pending on GitHub) tinted by the attention class (**red = act, yellow = agent working, grey = parked, neutral = resting**; tooltip spells it out). The Attention column is removed; the Sandbox chip keeps the machine-level detail (running/paused/failed). Attention remains a backend field driving sorting, UP NEXT and the badges — it just stopped being a column. One less cognitive hop, nothing lost.


barney-s:
Also per review: columns reordered on the same principle — **GitHub facts left (# · Title · Assignee · Age), repo-agent state right (Status · Sandbox · actions)**. Age moved left too since it's GitHub's updatedAt. The table now reads as 'the work | what the agent is doing about it'.

PR #1440, opened 2026-09-18
fix(envd): prevent false-positive quota kills on transient 429 retries

sdowell:
Previously, RunTaskResilient evaluated geminitokens.IsFatalQuotaError on isolated byte deltas from each tail -c poll tick, and treated generic googleapis 429 error text ("exceeded your current quota", "check your plan and billing details") as unconditionally fatal before checking for "Retrying with backoff". Because generativelanguage.googleapis.com includes that boilerplate text on all HTTP 429 responses (including transient RPM/TPM rate limits), tasks actively retrying with backoff were immediately SIGKILLed.

- Move generic 429 boilerplate strings to ambiguous quota indicators so they are deferred when accompanied by "Retrying with backoff", while keeping explicit RPD metrics ("requests per day", "Generate requests per day"), "Max retries exceeded", and suspended keys as unambiguous fatal errors.
- Introduce QuotaStreamTracker with an 8KB sliding window and 1-poll grace period so retry indicators and error payloads split across poll boundaries are evaluated together without window pollution.


factory/pkg/geminitokens/geminitokens.go
```diff
@@ -391,36 +391,70 @@ func ExtractAPIKeyFromError(data []byte) string {
...
+		if strings.Contains(str, sub) {
+			return true
+		}
+	}
+	return false
+}
+
+// ContainsRetryBackoff checks if the output contains a backoff retry indicator.
+func ContainsRetryBackoff(data []byte) bool {
+	return strings.Contains(string(data), retryBackoffMarker)
```
barney-s: `ContainsRetryBackoff` converts `data` to a `string` before calling `strings.Contains`. Since this check is performed on every poll tick, consider using `bytes.Contains(data, []byte(retryBackoffMarker))` to avoid string allocation overhead.

factory/pkg/geminitokens/geminitokens.go
```diff
@@ -391,36 +391,70 @@ func ExtractAPIKeyFromError(data []byte) string {
...
+// Note: Generic 429 messages from generativelanguage.googleapis.com ("You exceeded your
+// current quota, please check your plan and billing details") are returned on transient
+// RPM/TPM rate limits as well, so they are classified as ambiguous and deferred when
+// accompanied by "Retrying with backoff".
+func IsUnambiguousFatalQuotaError(data []byte) bool {
 	if IsSuspendedKeyError(data) {
 		return true
 	}
 
 	str := string(data)
```
barney-s: In `IsUnambiguousFatalQuotaError`, `data` is converted to a string `str := string(data)`. This allocation can be avoided by performing raw byte slice matching or using `bytes.Contains` directly for the target substrings.

factory/pkg/envd/client.go
```diff
@@ -423,6 +423,7 @@ func (c *Client) RunTaskResilient(ctx context.Context, cmdStr string, envs map[s
 	// 2. Tailing & status loop
 	var offset int64
 	var lastModelTried string
+	quotaTracker := geminitokens.NewQuotaStreamTracker()
```
barney-s, line 426: The local `quotaTracker` maintains an in-memory window bounded at 8KB via `trimWindow()`. Please verify if extremely large raw log chunks read in a single polling interval can cause a transient memory allocation spike before the deferred trimming occurs, and if a pre-append check on the chunk size is warranted.

barney-s (commented):
The overall design is sound, and the implementation is supported by deterministic unit tests covering split log boundaries and window pollution. Minor optimizations are suggested regarding string allocations in key helper functions to improve runtime performance.

factory/pkg/geminitokens/geminitokens.go
sdowell: Good catch, done. Switched to `bytes.Contains` against a package-level `retryBackoffMarkerBytes` so the marker itself isn't re-allocated per call either.

I applied the same treatment to the rest of the matchers and to `findUncoveredAmbiguousError`, which was doing `string(t.window)` before its `strings.Index` scan — that was actually the largest copy of the three, since it ran over the whole rolling window on every tick.

Added `BenchmarkQuotaStreamTrackerObservePoll` (full 8KB window, one delta line per tick) to keep this honest:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| before | 41229 | 28534 | 3 |
| after | 33332 | 118 | 0 |

The steady-state poll path is now allocation-free.

factory/pkg/geminitokens/geminitokens.go
sdowell: Done — the fatal markers are now a package-level `[][]byte` matched with `bytes.Contains`, so there's no string conversion:

```go
func IsUnambiguousFatalQuotaError(data []byte) bool {
	if IsSuspendedKeyError(data) {
		return true
	}
	for _, marker := range unambiguousFatalMarkers {
		if bytes.Contains(data, marker) {
			return true
		}
	}
	return false
}
```

Also fixed `IsSuspendedKeyError`, which this delegates to — it was copying the window a second time on every call, so this function was doing 2x the allocation you flagged. Benchmark numbers are in the thread above.

factory/pkg/envd/client.go
sdowell: Verified, and it was a real gap — thanks for pushing on it.

`BuildTailLogCmd` is an unbounded `tail -c +<offset>`, and the realistic worst case is reattaching to an already-running task: `offset` starts at 0, so the first poll returns the entire existing log in one chunk. The old `append` + deferred `trimWindow()` meant we transiently held `len(window) + len(chunk)` on top of the chunk itself.

Added the pre-append bound you suggested. `appendChunk()` detects chunks larger than `maxWindowSize`, scans them in place (free now that matching is on `[]byte`), and retains only the trailing 8KB by copying into the existing backing array:

```go
func (t *QuotaStreamTracker) appendChunk(newData []byte) (analysisBuf []byte, baseOffset int64) {
	if len(newData) > t.maxWindowSize {
		chunkStart := t.totalOffset
		t.totalOffset += int64(len(newData))
		// Reuse the existing backing array; it is already capped at maxWindowSize.
		t.window = append(t.window[:0], newData[len(newData)-t.maxWindowSize:]...)
		t.windowStartOffset = t.totalOffset - int64(t.maxWindowSize)
		return newData, chunkStart
	}
	...
}
```

Detection semantics are unchanged: the full oversized chunk is still scanned for both unambiguous fatal markers and uncovered ambiguous 429s, and the returned base offset keeps the grace-period stream offsets consistent across the truncation. Three new subtests cover it — bounded window, fatal marker sitting in the discarded prefix, and a grace period spanning an oversized chunk.

One residual I deliberately left alone: the chunk is already fully materialized in the `bytes.Buffer` here in `RunTaskResilient` before the tracker ever sees it, and it's echoed to stdout verbatim. Capping that read would drop the backlog replay on reattach, so I kept it as-is — the tracker just no longer amplifies the spike, which is the part this PR owns. Happy to bound the tail read in a follow-up if you think the reattach replay is expendable.

factory/pkg/envd/client.go
```diff
@@ -443,6 +444,30 @@ func (c *Client) RunTaskResilient(ctx context.Context, cmdStr string, envs map[s
...
+		var finalData []byte
+		if err := c.Exec(loopCtx, finalTailCmd, "/workspaces", nil, nil, &finalBuf, nil); err == nil {
+			finalData = finalBuf.Bytes()
+			if len(finalData) > 0 {
+				_, _ = os.Stdout.Write(finalData)
+				offset += int64(len(finalData))
+			}
+		}
+		isFatalQuota := quotaTracker.ObserveFinal(finalData, processFailed)
+		if model := extractModelFromLogs(quotaTracker.Window()); model != "" {
```
barney-s: When `newData` is oversized (larger than 8KB), `quotaTracker.ObserveFinal` scans the entire chunk for fatal errors but only retains the trailing 8KB in the sliding window. Because `extractModelFromLogs` is then called on `quotaTracker.Window()`, any "Trying model: ..." entry located in the earlier part of the oversized chunk will be missed. This leaves `lastModelTried` empty, causing the system to fall back to marking all models as quota-exceeded rather than the specific model that failed.

To fix this, check `finalData` first (or in addition to the window):
```go
if model := extractModelFromLogs(finalData); model != "" {
	lastModelTried = model
} else if model := extractModelFromLogs(quotaTracker.Window()); model != "" {
	lastModelTried = model
}
```

factory/pkg/envd/client.go
```diff
@@ -473,21 +498,21 @@ func (c *Client) RunTaskResilient(ctx context.Context, cmdStr string, envs map[s
...
-						defer killCancel()
-						killCmd := BuildQuotaKillCmd(taskFiles.PIDFile, taskFiles.StartTimeFile, taskFiles.ExitCodeFile)
-						_ = c.Exec(killCtx, killCmd, "/workspaces", nil, nil, nil, nil)
-						return handleQuotaOrSuspensionError(newData, lastModelTried, envs)
-					} else if geminitokens.IsTransientRateLimit(newData) {
-						klog.V(2).Infof("Transient rate limit (RPM/TPM) detected in task output; allowing CLI to retry with backoff...")
-					}
+				}
+				isFatal, isTransient := quotaTracker.ObservePoll(newData)
+				if model := extractModelFromLogs(quotaTracker.Window()); model != "" {
```
barney-s: Similarly to `flushAndCheckFinalQuota`, if `newData` is oversized, calling `extractModelFromLogs` on `quotaTracker.Window()` instead of `newData` will miss any model name logs that were printed in the earlier parts of the oversized chunk.

To preserve accurate model tracking, scan `newData` directly first:
```go
if model := extractModelFromLogs(newData); model != "" {
	lastModelTried = model
} else if model := extractModelFromLogs(quotaTracker.Window()); model != "" {
	lastModelTried = model
}
```

factory/pkg/geminitokens/geminitokens.go
```diff
@@ -391,52 +408,309 @@ func ExtractAPIKeyFromError(data []byte) string {
...
+
+// Window returns the current rolling log buffer for extracting metadata such as API keys or models.
+func (t *QuotaStreamTracker) Window() []byte {
+	return t.window
+}
+
+func (t *QuotaStreamTracker) trimWindow() {
+	if len(t.window) > t.maxWindowSize {
+		excess := len(t.window) - t.maxWindowSize
+		t.window = t.window[excess:]
```
barney-s: Slicing `t.window` using `t.window = t.window[excess:]` shifts the slice pointer forward but retains the underlying array with a reduced capacity. As subsequent appends are made, they write into the remaining capacity. Once capacity is exhausted, Go will allocate a new backing array and copy elements over. This leads to periodic heap allocations and garbage collection overhead during continuous log polling.

To make the sliding window completely allocation-free once it reaches maximum size, shift the elements to the beginning of the slice using `copy`:
```go
copy(t.window, t.window[excess:])
t.window = t.window[:t.maxWindowSize]
```

factory/pkg/geminitokens/geminitokens.go
```diff
@@ -391,52 +408,309 @@ func ExtractAPIKeyFromError(data []byte) string {
...
+			braceDepth++
+		} else if ch == '}' {
+			if braceDepth > 0 {
+				braceDepth--
+			}
+		} else if ch == '\n' && braceDepth == 0 {
+			nextLine := buf[i+1:]
+			if bytes.HasPrefix(nextLine, spacePrefix) ||
+				bytes.HasPrefix(nextLine, tabPrefix) ||
+				bytes.HasPrefix(nextLine, apiErrorPrefix) ||
```
barney-s: The manual increment `i++` inside this conditional block is crucial because the loop is defined as `for i < maxEnd` and does not have a post-iteration statement. However, this is a "hidden trap" for maintainers. If this loop is ever refactored to a standard three-clause `for i := retryStart; i < maxEnd; i++`, the `continue` statement will trigger the post-iteration `i++`, causing a silent double-increment bug that skips the first character of the next line and breaks brace depth tracking.

Please add a clear warning comment explaining this structural dependency, or refactor the loop to a standard `for ; i < maxEnd; i++` structure with all increments handled cleanly.

barney-s (commented):
Few opportunities for improvement:
1. Memory Efficiency: Slicing the sliding window causes amortized periodic allocations. Using `copy`-based shifting will make it completely allocation-free.
2. Model Extraction: Oversized chunks may discard the model name print, resulting in inaccurate model quota tracking.
3. Loop Maintenance: The manual loop increment inside `computeRetryCoverEnd` is correct but represents a subtle refactoring hazard that should be warned about or refactored.

factory/pkg/geminitokens/geminitokens.go
sdowell: Great catch, and it explains the loose end from the last round: the `118 B/op` still showing in the benchmark I posted was exactly this amortized reallocation — the window walking forward through its backing array until the remaining capacity ran out.

Switched to the `copy`-based shift:

```go
copy(t.window, t.window[excess:])
t.window = t.window[:t.maxWindowSize]
t.windowStartOffset += int64(excess)
```

Benchmark is now clean:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| original | 41229 | 28534 | 3 |
| after byte-matching | 33332 | 118 | 0 |
| after copy-shift | 33623 | 0 | 0 |

Also added a `testing.AllocsPerRun` assertion in the tracker tests so a future reslice regresses loudly instead of silently.

factory/pkg/geminitokens/geminitokens.go
sdowell: Agreed — rather than document the trap I removed it. Converted to a standard three-clause loop so `i` is incremented in exactly one place, and pulled the prefix checks out into `isPayloadContinuation`:

```go
for i := retryStart; i < maxEnd; i++ {
	switch ch := buf[i]; {
	case ch == '{':
		braceDepth++
	case ch == '}':
		if braceDepth > 0 {
			braceDepth--
		}
	case ch == '\n' && braceDepth == 0:
		if isPayloadContinuation(buf[i+1:]) {
			// Still inside a multi-line payload; keep scanning.
			continue
		}
		return i
	}
}
```

(`continue` inside a `switch` binds to the enclosing `for`, so the post statement still runs.) Behavior is unchanged — the retry-coverage and window-pollution tests all still pass.

factory/pkg/envd/client.go
sdowell: You're right, and this is a regression I introduced with `appendChunk` in the last round — before that, `trimWindow()` was deferred, so the window still held the entire chunk at the point `extractModelFromLogs` ran. Bounding the window at append time narrowed that view without my noticing the downstream consumers. Good catch.

Fixed, with one tweak to your suggestion: the same truncation also affects `handleQuotaOrSuspensionError(quotaTracker.Window(), ...)` a few lines below, which extracts the API key and decides suspension vs. quota — a truncated window there could downgrade a suspended key to a 4h quota block. Since the two buffers are always nested (ordinary poll → window ⊇ chunk, because the window also carries history; oversized chunk → chunk ⊇ window, because the window is just its tail), picking the longer one is lossless in both directions, so one helper covers every call site:

```go
func quotaEvidence(chunk []byte, window []byte) []byte {
	if len(chunk) > len(window) {
		return chunk
	}
	return window
}
```

```go
evidence := quotaEvidence(finalData, quotaTracker.Window())
if model := extractModelFromLogs(evidence); model != "" {
	lastModelTried = model
}
...
return handleQuotaOrSuspensionError(evidence, lastModelTried, envs)
```

Same outcome as your `if/else if` for the model case, and it repairs the key/suspension path too.

Added `factory/pkg/envd/client_test.go` with three cases: buffer selection, the oversized-chunk regression (it asserts the model really is absent from the retained window first, so the test can't silently stop reproducing the bug), and the small-chunk path where the window's history is what carries the model.

factory/pkg/envd/client.go
sdowell: Same fix applied here:

```go
isFatal, isTransient := quotaTracker.ObservePoll(newData)
evidence := quotaEvidence(newData, quotaTracker.Window())
if model := extractModelFromLogs(evidence); model != "" {
	lastModelTried = model
}
...
return handleQuotaOrSuspensionError(evidence, lastModelTried, envs)
```

`quotaEvidence` picks whichever of the raw chunk or the tracker window is the superset of the other, so the oversized case keeps the model name and the ordinary case keeps the history from earlier polls. Rationale and test coverage are in the `flushAndCheckFinalQuota` thread above.

barney-s (approved):

barney-s (approved):

PR #1450, opened 2026-09-18
ux(repo-agent): review rows show the PR author as a title chip

barney-s:
Review-tab rows (and review rows in Up Next) now lead the title with a `⦿ author` chip — same treatment as the assignee chip on issue rows (green when it's you, blue for others), so you can see whose work you're reviewing without opening the PR.

Scoped to the review group: on My PRs the author is always you, and issue rows already carry the assignee chip. `WorkItem.Author` added to the feed model, populated from the PR's user.


barney-s:
Added to this PR per live finding: clicks deferred by launch limits now render **queued** (waiting) instead of **starting** — with the default `maxActivePerUser: 2`, review clicks 3 and 4 sat on 'starting' while the controller logged `review deferred: executor at maxActivePerUser`. The feed now checks running sandboxes against the board's effective limit (min of the two knobs, with controller-default mirroring) and only calls a click 'starting' when it can actually launch. Test covers the third-click-at-capacity case.

PR #1452, opened 2026-09-18
fix(repo-agent): finished sandboxes stop holding launch slots

barney-s:
**Stacked on #1451 (which stacks on #1450) — merge in order.**

Live-observed: with the limit at 2, two *finished* reviews (pending reviews already posted on GitHub) kept holding both launch slots while two clicked reviews sat queued — a completed sandbox idles up to `idleMinutes` (default 60) before `pauseFinished` parks it, and the whole time it counted as active.

The limit caps concurrent **agent work**; an idle pod waiting out its pause window is not work. New `sandboxSettled` predicate: last task `Completed`/`Failed` and nothing newer waking it (`unpaused-at`, re-review/re-fix markers, plan feedback — all must be older than the completion stamp). Settled sandboxes are excluded from the controller's `activeCount` and mirrored in the feed's `queued` computation, so queued clicks now launch the moment the previous run *finishes*, not when it gets paused an hour later.

Waking sandboxes still count (their marker/wake stamp is newer than completion), so the wake window can't over-launch; provisioning sandboxes (no completion stamp) count too.

Tests: settled occupant frees the slot / running occupant still blocks, plus the predicate's edge cases.


barney-s:
Reworked per review discussion: **maxActive stays a pod cap.** Settled sandboxes still don't count as *work* (so queued clicks launch the moment a run finishes), but they no longer escape the pod budget either — `reclaimPodSlots` pauses the oldest settled sandboxes immediately whenever running pods exceed `maxActive`, yielding the keep-warm window early under pressure. Working and waking sandboxes are never evicted.

Net behavior with 100 auto-reviews at `maxActive: 5`: never more than ~5 pods, continuous throughput, keep-warm honored only when there's headroom. One knob, both meanings.

barney-s:
Closing per design discussion: **maxActive counting every running pod is the better design.** The finished pod holding its slot until the pause isn't a bug — it's the churn throttle. With settled-pod exclusion (or eviction), 100 auto-intake reviews would burn tokens at full maxActive concurrency continuously; with plain pod counting, the pause window naturally rate-limits intake churn while still capping the cluster footprint. One knob, and the slot-hold latency problem is addressed where it actually hurt: #1453 drops the *review* pause window to 5 minutes, so clicked reviews queue for minutes, not an hour. #1453 is rebased directly onto main and carries the rationale in its commit message.

PR #1453, opened 2026-09-18
feat(repo-agent): finished review sandboxes pause after 5 minutes

barney-s:
**Top of the stack: #1450 → #1451 → #1452 → this.**

`idleMinutes` (default 60) applies uniformly per board, but the keep-warm argument only holds for fix/plan sandboxes — their checkout and toolchain pay off across iterations, retries, and plan→fix continuity. A review sandbox is spent the moment its pending review lands on GitHub, and a re-review wakes it from paused anyway (live-measured ~4 min warm wake).

`pauseFinished` now uses `min(5m, idleMinutes)` for sandboxes whose last task type is `review`; every other type keeps the board knob. No CRD change — the per-type window is a constant, not another knob.

Test: at 10 minutes idle, a finished review sandbox pauses while a finished fix sandbox stays up.


barney-s:
Closing per design discussion: keeping one predictable knob. `idleMinutes` applies uniformly to every task type — no hardwired per-type pause windows. If review churn/queue latency needs tuning on a given board, lower that board's `spec.sandbox.idleMinutes`; the throttle stays explicit and visible in the spec rather than split across constants.

PR #1457, opened 2026-09-18
feat(repo-agent): reject triage; plan verbs as row buttons; hand-editable plans

barney-s:
Three draft-loop gaps from live testing:

1. **Reject triage** (row button next to Publish). Clears the draft breadcrumbs, tombstones the sandbox (`triage-rejected-at`), parks the pod, and the row reads **fully reset**: no sandbox chip, stage back to untriaged, **Triage / Plan / Fix** offered again. Consent semantics: auto-triage never redoes work a human threw away; a stale runner result can't resurrect the draft (and its backoff can't block a re-click); a fresh **Triage click re-arms**. The mailbox triage claim now stands until a draft is stored so the re-click can't be swallowed by the leftover sandbox.
2. **Plan-ready verbs surface as row buttons** — Approve & Fix and Reject plan sit with every other action instead of hiding in the fold-out. The sub-row keeps the plan text, the Refine feedback box, and now…
3. **Hand-editable plans**: Edit → textarea → Save (`PUT plan-draft`) for quick refinements without a full agent round. A manual edit bumps `planned-at` so a stale feedback stamp can't fire a refine that overwrites your edit.

Also fixes a regression I introduced in #1448: attaching the triage sandbox to resting rows tripped the `!item.sandbox` guard and **hid Fix/Plan on triage-ready rows** (why the row only offered 'Publish triage'). Stage is the guard now, and finished-empty (rejected) triage sandboxes stop attaching so reset rows read reset.

Tests: reject end-to-end (breadcrumbs cleared, tombstone stamped, feed shows the reset row), plus the controller tombstone/re-click semantics ride the existing triage suite.


barney-s:
Added per follow-up: published triage now renders as a green **Triaged ✓** state chip (was a blank status) — same done-state pattern as Reviewed ✓. The chip toggles the published suggestions (▾), the rail 'Suggestions' button retires for that stage, and the row's verbs are just **Plan | Fix**. Lifecycle reads: Triage → Triage ready ▾ (Publish | Reject | Fix | Plan) → Triaged ✓ (Plan | Fix).

barney-s:
Follow-up: **triage-ready now offers only Publish triage | Reject triage** — the same one-decision-at-a-time rule plan-ready follows. Lifecycle: `Triage → Triage ready ▾ (Publish | Reject) → Triaged ✓ (Plan | Fix)` or `→ rejected (Triage | Plan | Fix)`. Prevents the buried-draft state where a Plan started on top of a pending triage draft would hide it unpublished forever.

PR #1459, opened 2026-09-18
refactor(factory): only the task queue touches task files on disk

sdowell:
The PR and issue scanners walked the processed/ directory themselves to recover what had already been done — when an issue was last worked on, and which commit of a pull request was last reviewed, investigated or rebased. That made them second readers of state the queue owns, each carrying its own copy of the task file layout and of the rule that a file's modtime stands in for a missing completedAt. A directory read is also frozen at first use: anything that finished later was invisible.

- TaskQueueManager gains GetProcessedTask and ListProcessedTasks, both returning copies — a caller that could mutate what it read back would be editing queue state without the queue knowing.
- LoadFromDisk dates a recovered processed task by its file's modtime when it recorded no completedAt, so that rule now lives with the files and no reader has to stat them.
- EnsureDirs moves queue directory creation off the watcher.
- ProcessedDir is gone from both scanner Configs; neither knows where task files live.

Two behaviour changes fall out: the scanners now also see tasks that finished since start-up, and a processed file that fails to parse is skipped rather than folded in by timestamp alone.


factory/pkg/commands/watch/concurrency/queue_manager.go
```diff
@@ -479,6 +525,50 @@ func (m *TaskQueueManager) HasActivePRTask(prNumber int) bool {
...
+func (m *TaskQueueManager) GetProcessedTask(filename string) *api.QueueTask {
+	filename = filepath.Base(filename)
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	t, ok := m.processed[filename]
+	if !ok {
+		return nil
+	}
+	taskCopy := *t
```
barney-s: The dereference copy `taskCopy := *t` performs a shallow copy of the `api.QueueTask` struct. Since `QueueTask` contains a slice field (`Instructions []string`), the slice descriptor is copied by value, but it continues to reference the same underlying array as the queue's original task object.

If a caller of `GetProcessedTask` mutates the elements of `Instructions`, it will modify the task queue's internal state. To guarantee complete state isolation and prevent potential race conditions or state contamination, perform a deep copy of the `Instructions` slice or implement a `.Clone()` method on `api.QueueTask` to return a fully insulated copy.

factory/pkg/commands/watch/concurrency/queue_manager.go
```diff
@@ -479,6 +525,50 @@ func (m *TaskQueueManager) HasActivePRTask(prNumber int) bool {
...
+// one when it finishes a task, and dates the files it recovers by their own
+// timestamp. Callers can therefore order the processed set without going back
+// to the filesystem for it.
+func (m *TaskQueueManager) ListProcessedTasks() map[string]*api.QueueTask {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	snapshot := make(map[string]*api.QueueTask, len(m.processed))
+	for filename, t := range m.processed {
+		taskCopy := *t
```
barney-s: The dereference copy `taskCopy := *t` here also performs a shallow copy of each task inside the loop. Because `QueueTask` contains a slice field (`Instructions []string`), the underlying backing array remains shared with the task queue's original objects.

To ensure complete isolation of the returned task snapshot and protect the queue's internal state from mutations by callers, perform a deep copy of the `Instructions` slice or use a deep-copy `.Clone()` method on the task.

barney-s (commented):
The core feedback focuses on the task-copying implementation within `GetProcessedTask` and `ListProcessedTasks`. While the queue returns copies to prevent callers from mutating internal state, the dereference copy (`*t`) is a shallow copy. Because `api.QueueTask` contains a slice field (`Instructions []string`), the slice header is copied, but the underlying backing array remains shared. Any modification to slice elements or array values by a caller would mutate the queue's internal state.

factory/pkg/commands/watch/concurrency/queue_manager.go
sdowell: Good catch — the copy was shallow, so the returned task and the queue's task indexed the same `Instructions` backing array, and the isolation the doc comment claims did not hold.

Added `QueueTask.DeepCopy()`, which copies the struct and gives the copy its own instructions array, and routed this accessor through it.

Chasing it turned up the other half of the contract: `finishTask` was storing the caller's `*QueueTask` straight into `m.processed`, so the dispatcher kept a handle on finished state no matter what the readers returned, and `Enqueue`'s `*task` left the stored task sharing its instructions with whoever queued it. Both now go through `DeepCopy`, so the queue copies at every boundary in and out.

factory/pkg/commands/watch/concurrency/queue_manager.go
sdowell: Same fix here — the loop now builds the snapshot with `t.DeepCopy()`, so each entry owns its instructions array rather than sharing the queue's.

The `hands back copies` test now writes through `Instructions[0]` on both accessors and through the pointer handed to `CompleteTask`; it fails if `DeepCopy` goes back to a struct copy.

sdowell:
Pushed as `2ca20b4f`. Summary of the review round:

- `api.QueueTask` gains `DeepCopy()`: a struct copy shares the `Instructions` backing array, so the copies the accessors handed out were not actually insulated.
- `GetProcessedTask` and `ListProcessedTasks` return deep copies.
- Two more places had the same aliasing, on the write side: `finishTask` stored the caller's `*QueueTask` in the processed map, and `Enqueue` stored a struct copy that still shared instructions with the caller. Both copy now, so the queue owns its state at every boundary.
- `TestQueueTask_DeepCopy` covers the nil receiver, field copy, unshared array and nil-stays-nil; `TestProcessedTaskAccessors/hands back copies` now tampers through the slice and through the pointer passed to `CompleteTask`, and fails if the copy goes shallow again.

barney-s (approved):

PR #1461, opened 2026-09-18
fix(factory): overseer/review is inherited once, not kept sticky

sdowell:
Removing overseer/review from a PR did nothing. The next cycle syncReferencedIssueLabels copied it back from the parent issue, and shouldAutoReviewPR consulted the parent directly anyway, so the bot kept reviewing regardless. Opting out meant unlabelling the parent issue, which also opts out every sibling PR it spawns — a reviewer wanting to take over a single PR had no way to say so.

- Non-sticky labels: a general mechanism for labels whose absence on a PR is itself an instruction. They are inherited from the parent issue once and then left alone. nonStickyLabels is the registry; the review opt-in is its only member today, and every other inherited label still follows the parent as before.

- Removals are read back from the PR's own 'unlabeled' events (new ListIssueEvents), so a reclaimed label stays reclaimed across a restart and is honoured by a watcher that was not running when it came off. An unreadable log holds the label back rather than applying it: a cycle of delay beats resuming the fight. The in-memory memo caches only removals, so it cannot wrongly suppress an inheritance; it exists to stop a reclaimed PR from costing a lookup every cycle for the rest of its life.

- shouldAutoReviewPR reads only the PR's own labels. The parent-issue fallback made removing the label a no-op, which is exactly what a human removing it is trying not to be.

- syncReferencedIssueLabels now reflects added labels onto the in-memory issue. The stop-label re-check immediately after it was reading the pre-inheritance set and could never fire.


factory/pkg/github/events.go
```diff
@@ -0,0 +1,52 @@
...
+// pagination was cut short by maxIssueEventPages, in which case callers must
+// not treat the absence of an event as proof that it never happened.
+func (c *Client) ListIssueEvents(ctx context.Context, number int) ([]*githubv39.IssueEvent, bool, error) {
+	if !c.Ready() {
+		return nil, false, errNoClient
+	}
+
+	// Start non-nil so that a successful call never returns nil: callers use a
+	// nil log to mean "could not be determined".
+	all := []*githubv39.IssueEvent{}
```
barney-s: Declaring the slice with pre-allocated capacity `make([]*githubv39.IssueEvent, 0, issueEventPageSize)` instead of `[]*githubv39.IssueEvent{}` can improve performance slightly by avoiding slice re-allocation overhead as elements are appended.

sdowell: Done. Kept `make(..., 0, issueEventPageSize)` rather than a nil slice so the "nil log means could not be determined" contract still holds, and noted in the comment that one page of capacity is what the common case needs.

PR #1462, opened 2026-09-18
design(factory): multi-engine sandboxes — Claude Code alongside Gemini CLI

barney-s:
Design doc only (`factory/design/multi-engine.md`) — for review before any implementation.

**TL;DR**: the engine is one invocation line, one JSON field name, one API key, and one usage-stats shape — every factory contract (banners, plan files, task states, sandbox lifecycle) lives *around* the CLI and is untouched. The sandbox image already installs both CLIs.

- Selection: `RepoBoard spec.sandbox.engine: gemini|claude` → `factorycli --engine` → `ENGINE` task env; default `gemini` end to end, so landing it is a no-op until a board opts in.
- Scripts: the 9 task scripts' shared plumbing extracts into an embedded `lib.sh` first (no-behavior-change PR), then `runEngine` grows a two-line claude branch (`claude -p --dangerously-skip-permissions --output-format json`, `.result` vs `.response`).
- Auth: `ANTHROPIC_API_KEY` through the factory-user secret (repo-agent copies it from the member's existing `anthropic-api-key` secret), fail-fast when the selected engine's key is missing.
- Usage: claude cost/tokens map from its output JSON into the existing per-task usage files (token-daemon unchanged); transcript-based tool telemetry is a phase-4 adapter.
- Rollout: 4 independent PRs, then A/B two boards on one repo — one engine each — before touching any default.

Open questions at the end (claude model fallback order, spec mutability, per-verb engines later).


factory/design/multi-engine.md
```diff
@@ -0,0 +1,215 @@
+# Multi-engine sandboxes: Claude Code alongside Gemini CLI
+
+Status: proposed
+Owner: barney-s
+Scope: factory (primary), repo-agent (plumbing), overseer (untouched)
```
sdowell, line 5: Do you think we'll consider future work to add plumbing/configuration for overseer?

factory/design/multi-engine.md
```diff
@@ -0,0 +1,215 @@
...
+  by the existing `MODELS` env passthrough).
+
+`pkg/tasks/models.go` gains `ModelsForEngine(engine, geminiKey) string`.
+
+### Usage reporting
+
+v1: **degrade gracefully, don't block.** `record_usage` branches:
+
+- gemini: existing stats parse + `~/.gemini/tmp` transcript mining.
+- claude: Claude Code's JSON output carries `usage` and `total_cost_usd`;
```
sdowell, line 134: Instead of munging/adapting the data, maybe we can define a generic UsageReporting-type interface and have separate implementations for claude/gemini - wdyt? I suspect having provider-specific implementations will be easier to maintain long term

factory/design/multi-engine.md
```diff
@@ -0,0 +1,215 @@
...
+  structured review YAML — all produced by the *script* extracting the
+  model's final response and by prompt instructions.
+- Task-state stamping, sandbox lifecycle, envd, reattach, factorycli.
+- Prompt templates (`pkg/tasks/*.txt`) — plain text, engine-neutral.
+- The RepoBoard controller and UI — they consume factory's contracts.
+
+The engine surface is exactly: one invocation line + one JSON field name +
+one API key + one usage-stats shape.
+
+## Current state (verified touchpoints)
```
sdowell, line 40: I suspect there may also be some of the (brittle 😿) gemini-specific logic that will need an equivalent for claude as well (e.g. the detection for quota exhaustion / rate-limiting). Is that accounted for?

factory/design/multi-engine.md
barney-s: agreed

factory/design/multi-engine.md
barney-s: no we need to discover it as we enable usage and run into issues. But will flag this.

factory/design/multi-engine.md
barney-s: wanted to limit impact on kcc work

sdowell: Oh definitely, agreed - I'm on board for keeping it out of scope for short term. Was mostly just curious what the long term vision is - but that's probably hard to predict 😄

factory/design/multi-engine.md
sdowell: Not sure if that's clear in the implementation plan, might be worth spelling out

sdowell (approved):

PR #1465, opened 2026-09-18
feat(repo-agent): All — the cross-board Up Next inbox

barney-s:
The cross-board inbox from today's design discussion: an **All** entry in the board selector (shown when you have more than one board, and the **default landing**) that renders a single aggregated **Up Next** — every `needs-you` row across all your boards, oldest demand first.

- **Rows act inline against their own board**: each row is tagged with its board, so Publish/Approve/Retry/Finalize post to the right endpoints without navigating. Read-only roles apply per row's board.
- **The board chip on each row is clickable** — jump into that board for fold-outs and full context.
- **Just Up Next, deliberately**: a cross-repo triage queue or PR list mixes hats; an inbox of things waiting on *you* is one hat. The All tab shows the red total badge (sum of per-board needs-you).
- **Zero new API / zero extra quota**: client-side aggregation over the #1463 server caches — N cached GETs answering in milliseconds; per-board `spec.view.labels` filters are inherited for free (server-side), fluid per-board UI filters deliberately don't apply.
- Empty state: `✓ Nothing needs you anywhere.` — the best sentence the tool can say.

Board-specific chrome (gear, delete, group tabs, view chips) hides on All since it has no single board.


barney-s:
Added per follow-up: **within needs-you, bare `Review requested` rows sort last** — finished agent work awaiting a verdict (Triage ready, Plan ready, Pending on GitHub, failures with Retry) comes first. Applied in the server sort (so every per-board Up Next inherits it) and mirrored in the All inbox's client sort. Test pins the ordering: a triage-ready draft outranks a newer bare review request.

PR #1470, opened 2026-09-19
fix(watch): a failed sandbox probe is not a verdict

sdowell:
Startup recovery asks the cluster what became of each task left in processing, and treated a question it could not get an answer to as an answer. recoverTask logged the error from IsTaskRunning or IsTaskCompleted and fell through to the requeue path, so a transient API failure was indistinguishable from a missing sandbox and the task was run again — work that may well have already finished.

The adoption monitor had the same shape. A failed completion probe left `completed` false, so it recorded a successful task as Failed and resolved the review comments it had acknowledged with a confused reaction.

Recovery now classifies the sandbox explicitly — running, completed, gone, or unknown — retrying a probe that fails. Unknown is no longer collapsed into gone: the task stays in processing for a later run to triage, because recovery is the only thing that inspects it and a wrong guess decides its fate. The monitor shares the same probe and simply keeps polling, leaving its timeout as the only thing that gives up on a sandbox that will not answer.


barney-s (approved):
lgtm

PR #1473, opened 2026-09-19
fix(factory): re-invocations reattach to in-flight tasks instead of duplicating

barney-s:
Live failure: a controller restart mid-plan left **4 parallel plan runs** in one sandbox (fix-…-1122) — each a full token spend. Root cause: `RunTaskResilient` already has the full watch-style reattach discipline (alive-pid check, completed check, tail-and-resume), but it keys on the task directory, and **every invocation mints a fresh timestamped dir** — so cross-invocation reattach never engaged, and each relaunch nohup'd a duplicate beside the live run.

**`envd.ResolveTaskDir(prefix)`**: if the newest `/workspaces/tasks/<prefix>-*` dir has a live pid and no `exit_code`, reuse it (skipping the prompt/script rewrite); `RunTaskResilient` then takes its existing reattach path — tails the in-flight run, prints the banner output at completion, stamps the final task state. A restarted controller re-invokes and simply *picks the run back up*: no duplicate tokens, harvest and task-state stamping intact (the annotation-orphan problem goes away too, since the reattached invocation lives to stamp Completed).

Applied to the four board-driven commands (plan, triage, review, fix). The `pr.*` flows (investigate/address/iterate/adopt) and `agent create` can adopt the same one-line helper as a follow-up.


barney-s:
Closing in favor of #1475 (runner-side dispatchTask discipline): same bug, zero factory delta, and the skip/adopt mechanism is generic across all board-driven commands (plan/triage adopt orphaned results outright; fix/review correct the record and relaunch cleanly). If end-to-end recovery of orphaned review/fix runs (re-running the host-side post step against the same task) ever becomes worth a factory change, this branch is the reference implementation.

PR #1474, opened 2026-09-19
perf(repo-agent): controller GitHub reads use conditional requests

barney-s:
Companion to #1463's API-side ETags: the controller's once-a-minute reconcile listings (auto-tier discovery, requested-review scans, settle passes, launch guards) now go through a shared `httpcache` transport. GitHub answers unchanged resources with **304 — zero rate-limit quota** — so steady-state reconciling of quiet repos is nearly free, and quota is spent only when something actually changed.

`memberGithubClient` now routes through the same cached builder as `githubClientFromToken` (one construction path). Per-token cache keying comes from GitHub's `Vary: Authorization`.

With this, both pollers (browser-driven feed + controller reconcile) are conditional-request-backed — the quota concern from the polling discussion is structurally addressed.


barney-s:
**'Are we sure the ETags work?' — good question, the proof-test said no-with-an-asterisk.** The conditional flow itself works (first GET caches, second sends `If-None-Match`, 304 → cached body, quota-free per GitHub's documented behavior). But the test caught a real bug: with oauth2 nested *inside* the cache transport, httpcache never saw the `Authorization` header, so `Vary: Authorization` matched vacuously — **cached bodies shared across tokens**. That inversion also shipped in #1463's API client, where `ListReviews` responses carry viewer-private pending reviews (and two members do share the substrate repo today).

Fixed in both clients (auth outside, cache inside) with a test pinning the whole contract: cache→conditional→304-served-body→cross-token isolation. This PR now also carries the #1463 correction.

Remaining honest caveat: GitHub responses carry `max-age=60`, so some back-to-back reconciles are served from local cache with zero request — even cheaper, at worst ~60s staler; automation latency impact ≤1 reconcile.

PR #1478, opened 2026-09-19
fix(repo-agent): the Overseer tab terminal moves to the tmux transport

barney-s:
The 'keeps disconnecting' Overseer-tab terminal, fixed the way #1477 fixed the card's: same transport, shared core.

- **Backend**: the tab's existing route (`/api/terminal/:namespace/:name`, same authz) now runs the namespace-parameterized tmux/exec core from `handlers_sandbox_terminal.go`. The ws→exec→`factory sshd` tunnel — three layers, zero keepalives, sessions that evaporated on drop — is retired (`factory sshd` itself untouched, just no longer called).
- **Frontend**: `Terminal.js` keeps its component interface and its frame protocol (which the shared core now speaks verbatim), gaining the reconnect-with-backoff loop and an honest status line. Because the session lives in tmux, reconnects land back in the same shell with state intact — and since both UIs attach `tmux -As board`, an admin's Overseer shell and the member's card shell are **two views of one session** (tmux multi-attach, by design).
- **Cleanup**: xterm unifies on the npm package the Overseer terminal already used; #1477's vendored `public/vendor` copies and index.html script tags are removed.

Manual test: Overseer tab → sandbox → terminal → run `top` → toggle devtools offline for 30s → reconnects into the same `top`.


barney-s:
Live finding while testing: connecting to a **ContainerCreating** pod (cold pull — `fix-…-1431`) looped `[connection to sandbox ended]` because the pod exists but accepts no exec yet. The handler now checks pod readiness and reports *'sandbox pod is starting (Pending) — reconnecting until it's ready…'* — the frontend's existing backoff converges the moment the container is up. (tmux presence in overseer pods verified — the long-lived overseer pod has it.)

